PR #20, opened 2026-04-13
fix: bundle prepare fallback to stopped snapshots

bloodcarter:
## Summary
- make `hcom bundle prepare --for <stopped-agent>` fall back to stopped life-event snapshots when the live `instances` row no longer exists
- keep the existing live-instance path unchanged
- add a regression test covering a stopped agent with a valid transcript file and session id

## Root cause
`cmd_bundle_prepare()` only looked up `transcript_path`, `tool`, and `session_id` from the live `instances` table. For stopped agents, those values may only exist in the latest stopped life-event snapshot, so `transcript.range` stayed null even when the stopped agent still had a valid transcript file and exchange history.

## Testing
- `cargo test test_lookup_bundle_transcript_source_falls_back_to_stopped_snapshot --quiet`
- `cargo test test_bundle_prepare_parse --quiet`


aannoo:
thanks

PR #23, opened 2026-04-19
fix(opencode-plugin): close TOCTOU race in deliverPendingToIdle

mmkzer0:
## Summary
This fixes duplicate injected hcom turns in the OpenCode plugin by closing a TOCTOU race in `deliverPendingToIdle()`.

The issue became reproducible under multi-agent orchestration, where wake events happen frequently and near-simultaneously.

## Root cause
`deliverPendingToIdle()` can be entered from two independent wake paths:
- TCP notify wake
- idle-status wake (`session.status` -> `idle`)

Both paths could pass the `pendingAckId === null` check before either caller set it (first `await` boundary), causing duplicate `opencode-read`/`promptAsync` delivery of the same payload.

## Fix
- Add synchronous `deliveryInFlight` guard flag checked/set before the first `await`
- Early-return concurrent callers when a delivery is already running
- Reset guard in `finally`

`pendingAckId` remains in place as the second serialization layer between injection and transform-hook ack.

## Validation
- Reproduced prior duplicate-delivery behavior under frequent wake conditions
- Two-turn injection tests (`FIRST_OK` / `SECOND_OK`) now show single delivery per message
- Real OpenCode usage confirms duplicate turn injection is no longer present
- Repository test suite passes (`cargo test -q`)

## Notes
This PR is intentionally scoped to the delivery race fix only. The agent/model persistence follow-up is prepared as a separate stacked branch/PR.

### DIFF
--- a/src/opencode_plugin/hcom.ts
+++ b/src/opencode_plugin/hcom.ts
@@ -84,8 +84,7 @@ export const HcomPlugin: Plugin = async ({ client, $ }) => {
   //     next wake event.
   //   pendingAckId — set after messages are read, cleared by transform.
   //     Prevents re-delivery while a prior injection is still being processed.
-  //     If promptAsync throws synchronously after pendingAckId is set, the id
-  //     remains until the next transform fires or the session resets.
+  //     If promptAsync fails to queue, pendingAckId is cleared immediately.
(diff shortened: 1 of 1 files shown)


aannoo:
thanks

PR #26, opened 2026-04-22
Fix headless PTY background launches on macOS

mmkzer0:
# Fix headless PTY background launches on macOS
...oh man, this was an odyssey.

## Summary
Fixes the macOS `--headless --go` regressions on the PTY-backed launch path.

Validated end result:
- headless Codex works again
- headless OpenCode works again
- bare headless Claude without a prompt/task now fails fast with a clear error
- prompt/task-based headless Claude still works

## Motivation
Headless launches on the PTY path had regressed in several overlapping ways, which made this investigation a bit deceptive at first:

- background PTY launches could drop instance identity before `hcom pty` even started
- non-TTY stdin handling on macOS could prematurely tear down the PTY loop
- invalid/non-useful stdin could stay in the poll set and cause hot wakeups
- the non-blocking inject listener could keep waking spuriously on macOS after `WouldBlock`
- bare `hcom claude --headless` without a prompt/task looked like “headless PTY is broken” but is actually a separate unsupported detached-Claude path

User-facing symptoms included:
- Codex/OpenCode ending up in `launch_failed`
- missing `delivery.start` / `notify.registered`
- PTY spin / high CPU
- Claude launching and exiting immediately in unsupported bare headless mode

### Launcher / background PTY env
- Propagate `HCOM_INSTANCE_NAME` into the background PTY runner env
- Apply `tool_extra_env(tool)` on background PTY launches too

### PTY loop hardening
- Do not treat stdin EOF as terminal disconnect when stdin is not a TTY
- Fully remove stdin from the poll set when it should no longer be polled
- Drop stdin polling on `POLLNVAL`
- Ignore stdin `POLLHUP` for non-TTY headless PTY sessions
- Add one-iteration listener backoff after inject-listener `WouldBlock`

### PTY child identity
- Mark PTY child processes with `HCOM_LAUNCHED=1`

### Claude safety guard
- Reject bare `hcom claude --headless` without a prompt/task
- Keep prompt/task-based headless Claude behavior intact

## Notes / implementation details
- The main headless Codex/OpenCode fix was restoring `HCOM_INSTANCE_NAME` on the background PTY path so delivery can initialize for the correct instance.
- The PTY poll loop changes are specifically aimed at non-interactive stdin on macOS (`/dev/null`, closed pipes, invalid fds).
- The inject listener still stays non-blocking; the fix is not to change socket mode, but to back off for one iteration after `WouldBlock`.
- The `HCOM_LAUNCHED=1` PTY-child env fix ensures downstream hooks/tool integrations still recognize PTY-managed child sessions as hcom-launched.
- Bare headless Claude is intentionally not “fixed” into a PTY-backed live session model in this PR. The safe change here is explicit early validation.

### Automated
- `cargo test`
- Added regression coverage for:
  - background runner env propagation
  - Claude headless validation
  - PTY child launch env
  - inject listener accept helper behavior / timing resilience

### Manual
- `./target/release/hcom codex --headless --go`
  - launches successfully
  - transitions to `listening`
  - round-trip message delivery works
- `./target/release/hcom opencode --headless --go`
  - launches successfully
  - transitions to `listening`
  - round-trip message delivery works
- `./target/release/hcom claude --headless --go`
  - now fails fast with a clear prompt/task requirement
- `./target/release/hcom claude --headless --go 'Ping!'`
  - launches successfully
  - later message delivery round-trip works

### Active-binary PTY re-check
- Rebuilt release binary
- Launched fresh PTY-backed headless Codex from `./target/release/hcom`
- Observed normal `delivery.start` / `notify.registered`
- Observed no new recent `proxy.poll_spin_trace` / `POLLNVAL` log entries
- Focused process check showed normal CPU for PTY wrapper and child

### Final pre-PR hardening
- Updated PTY stdin handling so non-TTY `POLLHUP` disables stdin polling instead of breaking the PTY loop
- Made the inject accept regression test resilient to non-blocking localhost timing
- Reran focused PTY tests and full `cargo test` successfully

## Deferred follow-up
Not included in this PR:
- true PTY-backed headless Claude as a live background session model
- deeper investigation of Claude initial-prompt swallowing on detached launches
- any broader resume/fork cleanup around tool marker vars beyond the validated fixes here
(those are deferred follow-up items rather than blockers for restoring stable headless Codex/OpenCode PTY launches)

## Commit stack
- `fix(pty): don't treat stdin EOF as terminal disconnect when stdin is not a TTY`
- `fix(pty): mark PTY child processes as hcom-launched`
- `fix(pty): fully remove stdin from poll set when not a TTY`
- `fix(headless): restore stable PTY background launches`
- `fix(pty): tighten headless stdin and inject test handling`

### DIFF
--- a/src/commands/launch.rs
+++ b/src/commands/launch.rs
@@ -135,6 +135,8 @@ pub fn run(argv: &[String], flags: &GlobalFlags) -> Result<i32> {
     let system_prompt = hcom_flags.system_prompt;
     let initial_prompt = hcom_flags.initial_prompt;
 
+    validate_claude_headless_launch(&tool, background, &merged_args, initial_prompt.as_deref())?;
+
     // Open DB
     let db = HcomDb::open()?;
 
@@ -221,6 +223,29 @@ pub(crate) fn prepare_launch_execution(
     }
 
     (merged_args, background, use_pty)
+}
+
+fn validate_claude_headless_launch(
+    tool: &str,
+    background: bool,
+    merged_args: &[String],
+    initial_prompt: Option<&str>,
+) -> Result<()> {
+    if tool != "claude" || !background {
+        return Ok(());
+    }
+
+    let spec = claude_args::resolve_claude_args(Some(merged_args), None);
+    let has_cli_prompt = spec.positional_tokens.iter().any(|t| !t.trim().is_empty());
+    let has_hcom_prompt = initial_prompt.is_some_and(|p| !p.trim().is_empty());
+
+    if has_cli_prompt || has_hcom_prompt {
+        return Ok(());
+    }
+
+    bail!(
+        "Claude headless mode requires a prompt/task. Try `hcom claude --headless --hcom-prompt 'say hi in hcom'` or `hcom claude -p 'say hi in hcom'`."
+    )
 }
 
 pub(crate) fn launch_result_to_json(result: &LaunchResult) -> serde_json::Value {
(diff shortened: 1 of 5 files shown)


aannoo:
nice

PR #35, opened 2026-05-01
fix(opencode-plugin): allow --agent mode to register session binding

mmkzer0:
## Summary
OpenCode instances launched in headless `--agent` mode were invisible to `hcom hooks` and `hcom list` (showing up as PTY-only). This PR fixes the session registration bypass so they correctly create `session_bindings` rows.
*(Issue I encountered when codex should spawn headless opencode instance via hcom script when --agent was provided; env var worked).*

## Root cause
In `--agent` mode, OpenCode does **not** emit a `session.created` event. Execution starts directly at `chat.message` with the agent pre-configured.

`bindIdentity()` is the only path in the plugin that calls `hcom opencode-start --session-id ...` to create the session DB row. However, it was strictly guarded by:
```ts
if (process.env.HCOM_LAUNCHED !== "1") return
```

`HCOM_LAUNCHED` is injected by the PTY launcher, but *not* by the headless `--agent` path. This caused the guard to fire unconditionally. The fallback paths (`chat.message` and `transform`) hit the early return and skipped `opencode-start` entirely.

## Changes

**`HCOM_LAUNCHED` guard removal**
- Removed the strict `HCOM_LAUNCHED !== "1"` check in `bindIdentity()`.
- Safety is fully preserved by the internal `instanceName || bindingPromise` concurrency fence and the overarching `checkHcom()` binary verification gates.

**Fallback caller hardening**
- Added `!bindingPromise` pre-checks to `chat.message` and `experimental.chat.messages.transform`.
- This aligns the fallback callers with how the primary event handlers (`session.created`, `permission.asked`) already operate.

**`chat.message` unbound guard**
- Added an explicit `!instanceName && !sessionId` check.
- Previously, if `bindIdentity()` was called but failed (e.g. daemon absent), `isBoundSession(null)` would short-circuit to `true` and silently mutate agent/model state.
- Now, it emits a `WARN` (`plugin.chat_message_unbound`) and explicitly skips mutation.

**Regression Coverage**
- Added 4 new unit tests in `src/instance_binding.rs` covering the exact `--agent` mode binding invariants, lifecycle, and idempotency.

## Testing
- [x] `cargo test` (all 1473 pass, including new `instance_binding` coverage)
- [x] `cargo build --release`
- [x] Manual: Launch `hcom opencode --agent <name>` headless. Verified it successfully registers its session and is visible to `hcom hooks` and `hcom list`.

### DIFF
--- a/src/instance_binding.rs
+++ b/src/instance_binding.rs
@@ -1199,4 +1199,143 @@ mod tests {
 
         cleanup(path);
     }
+
+    // ── --agent mode regression tests ────────────────────────────────────────
+    //
+    // In `--agent` mode OpenCode skips `session.created` and fires `chat.message`
+    // first.  `bindIdentity()` is the only path that calls `opencode-start` and
+    // creates a `session_bindings` row.  The tests below verify the DB-level
+    // invariants that the plugin fix relies on.
+
+    /// Regression: calling `bind_session_to_process` without a prior
+    /// `session.created` (the --agent path) must still create a `session_bindings`
+    /// row.  Before the fix, `HCOM_LAUNCHED` blocked `bindIdentity` and no row
+    /// was ever inserted.
+    #[test]
+    fn test_agent_mode_bind_session_creates_session_binding() {
+        crate::config::Config::init();
+        let (db, path) = setup_test_db();
+        let now = now_epoch_i64();
+
+        // Simulate: instance placeholder created by the launcher (no session_id yet,
+        // matching the state before session.created would fire in normal mode).
+        let mut data = serde_json::Map::new();
+        data.insert("name".into(), serde_json::json!("agent-alpha"));
+        data.insert("status".into(), serde_json::json!("pending"));
+        data.insert("status_context".into(), serde_json::json!("new"));
+        data.insert("created_at".into(), serde_json::json!(now));
+        data.insert("tool".into(), serde_json::json!("opencode"));
+        db.save_instance_named("agent-alpha", &data).unwrap();
+        db.set_process_binding("proc-agent-1", "", "agent-alpha").unwrap();
+
+        // The plugin now calls this on the first chat.message (--agent path).
+        let result = bind_session_to_process(&db, "sess-agent-1", Some("proc-agent-1"));
+        assert_eq!(result, Some("agent-alpha".to_string()));
+
+        // session_bindings row must exist (was missing before the fix).
+        let binding = db.get_session_binding("sess-agent-1").unwrap();
+        assert_eq!(binding, Some("agent-alpha".to_string()));
+
+        // Instance must carry the session_id.
+        let inst = db.get_instance_full("agent-alpha").unwrap().unwrap();
+        assert_eq!(inst.session_id.as_deref(), Some("sess-agent-1"));
+
+        cleanup(path);
+    }
+
+    /// `bind_session_to_process` called twice with the same session_id must be
+    /// idempotent: same name returned, no duplicate session_bindings rows.
+    #[test]
+    fn test_agent_mode_bind_session_idempotent() {
+        crate::config::Config::init();
+        let (db, path) = setup_test_db();
+        let now = now_epoch_i64();
+
+        let mut data = serde_json::Map::new();
+        data.insert("name".into(), serde_json::json!("agent-beta"));
+        data.insert("status".into(), serde_json::json!("pending"));
+        data.insert("status_context".into(), serde_json::json!("new"));
+        data.insert("created_at".into(), serde_json::json!(now));
+        data.insert("tool".into(), serde_json::json!("opencode"));
+        db.save_instance_named("agent-beta", &data).unwrap();
+        db.set_process_binding("proc-agent-2", "", "agent-beta").unwrap();
+
+        let r1 = bind_session_to_process(&db, "sess-agent-2", Some("proc-agent-2"));
+        let r2 = bind_session_to_process(&db, "sess-agent-2", Some("proc-agent-2"));
+        assert_eq!(r1, Some("agent-beta".to_string()));
+        assert_eq!(r2, Some("agent-beta".to_string()));
+
+        // Exactly one session_bindings row.
+        let count: i64 = db
+            .conn()
+            .query_row(
+                "SELECT COUNT(*) FROM session_bindings WHERE session_id = ?",
+                rusqlite::params!["sess-agent-2"],
+                |row| row.get(0),
+            )
+            .unwrap();
+        assert_eq!(count, 1, "must have exactly one session_bindings row");
+
+        cleanup(path);
+    }
+
+    /// Full sequence: `initialize_instance_in_position_file` (tool=opencode, no
+    /// session_id) followed by `bind_session_to_process` — mirrors what happens
+    /// in the --agent path after the fix.
+    #[test]
+    fn test_agent_mode_initialize_then_bind_session() {
+        crate::config::Config::init();
+        let (db, path) = setup_test_db();
+
+        // Step 1: launcher creates the placeholder row (no session_id in --agent mode).
+        initialize_instance_in_position_file(
+            &db,
+            "agent-gamma",
+            None,  // no session_id yet
+            None, None, None, None,
+            Some("opencode"),
+            false, None, None, None, None, None,
+        );
+        db.set_process_binding("proc-agent-3", "", "agent-gamma").unwrap();
+
+        // Row exists, no session yet.
+        let before = db.get_instance_full("agent-gamma").unwrap().unwrap();
+        assert!(before.session_id.is_none(), "session_id should be absent before binding");
+
+        // Step 2: plugin calls bind on the first chat.message.
+        let result = bind_session_to_process(&db, "sess-agent-3", Some("proc-agent-3"));
+        assert_eq!(result, Some("agent-gamma".to_string()));
+
+        // session_bindings row created.
+        assert_eq!(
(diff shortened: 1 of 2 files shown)


aannoo:
Hey, I couldn't reproduce, it already was working fine for me - do you have more info?

AI:

# PR #35 review

target: `mmkzer0/hcom#35` (`fix/bindIdentity-headless`)
base: `main` @ `e0560a7a`
scope: `src/opencode_plugin/hcom.ts` env-guard swap + `chat.message` guard tightening + 4 Rust tests in `src/instance_binding.rs`

## live test

method: `hcom opencode --agent build --headless --hcom-prompt ...` on each branch, fresh agent, plugin log inspected.

| run | branch | instance | session | session_created | bound via | session_bindings row | reply |
|---|---|---|---|---|---|---|---|
| A | main | roge | ses_21c349dc7ffet3tMN0ymnBRMlg | yes | session.created path | yes | "live test ok" |
| B | PR | rila | ses_21c317bc6ffeQMiPZW3GGfXOoy | yes | session.created path | yes | "pr branch test ok" |

note: `plugin.session_created` log entry has no `instance` field (logs before `instanceName` is set). grepping by instance name omits it. earlier conclusion that `session.created` did not fire was an artifact of that filter and is retracted.

result: identical plugin log sequence on both branches. neither run exercised the `chat.message`-first / no-`session.created` path the PR claims to fix. PR is functionally inert for the tested invocation.

## env-guard swap analysis

change: `if (process.env.HCOM_LAUNCHED !== "1") return` → `if (!process.env.HCOM_PROCESS_ID) return` at `src/opencode_plugin/hcom.ts:295`.

evidence both vars are coupled in the standard launcher path:
- `src/launcher.rs:890` sets `HCOM_LAUNCHED=1`.
- `src/launcher.rs:912` sets `HCOM_PROCESS_ID=<uuid>`.
- same loop, no `--agent`-aware branching.
- `src/pty/mod.rs` `pty_child_env` also sets `HCOM_LAUNCHED=1`.
- launch scripts for both A and B exported both vars before exec of `hcom pty opencode --agent build --prompt ...`.
- `src/terminal.rs:716` unsets both together in post-exit cleanup.

implication: for any invocation that goes through `prepare_launch_execution`, `HCOM_LAUNCHED == "1"` iff `HCOM_PROCESS_ID` is set. swap is a no-op on this path.

PR description claim: "`HCOM_LAUNCHED` is injected by the PTY launcher, but not by the headless `--agent` path." not supported by the code on `main`. likely refers to a non-launcher path (e.g. codex subprocess spawning opencode via a wrapper that propagates one env var but not the other) — description does not specify.

## OpenCode behavior

PR claim: "in `--agent` mode, OpenCode does not emit a `session.created` event."

observed (current OpenCode in this env): `plugin.session_created` fired in both A and B with `--agent build`. OpenCode `run.ts` calls `sdk.session.create()` when no existing session is provided; subsequent `chat.message` carries `sessionID`. claim does not reproduce on the OpenCode version installed.

needed from author: OpenCode version/commit where `--agent` skips `session.created`.

## test coverage assessment

4 new tests in `src/instance_binding.rs`:
- `test_agent_mode_bind_session_creates_session_binding`
- `test_agent_mode_bind_session_idempotent`
- `test_agent_mode_initialize_then_bind_session`
- `test_agent_mode_no_process_id_no_session_binding`

all exercise the Rust DB function `bind_session_to_process`. none exercise:
- the TS plugin `bindIdentity` env-guard.
- the `chat.message` fallback bind path.
- the `HCOM_PROCESS_ID` set / `HCOM_LAUNCHED` unset asymmetry the PR claims to fix.

DB invariants tested were already correct; the load-bearing change is in `hcom.ts` and is uncovered. a future refactor of the guard could silently regress the target case.

## non-hcom session side effect

new branch at `src/opencode_plugin/hcom.ts:466-471` emits `WARN plugin.chat_message_unbound` per `chat.message` when `instanceName` is null.

triggers for users who:
- have `hcom` binary on `$PATH` (`checkHcom()` passes at `:449`).
- launch OpenCode without going through `hcom` launcher (`HCOM_PROCESS_ID` unset → `bindIdentity` early-returns at `:295`).
- have `~/.hcom/.tmp/logs/` populated (otherwise `appendFileSync` silently fails at `:62`).

result: every prompt appends a WARN line. WARN reason text "hcom absent or daemon error" is incorrect in this case (cause is "not launched by hcom").

remediation: gate the WARN on `process.env.HCOM_PROCESS_ID` so it fires only on genuine bind-attempt failures. preserves "hooks do nothing if you are not using hcom" guarantee.

## independent of env-guard issue

`chat.message` guard tightening at `:461-478` is correct on its own merits. old code's `isBoundSession(sid)` returned true when local `sessionId` was null (`!sessionId` short-circuit), allowing `currentAgent` / `currentModel` mutation pre-binding and from foreign sessions. new code requires `instanceName` set before mutation. retain regardless of env-guard outcome.

## defensive `!bindingPromise` checks

added at `:458` and `:495`. each is preceded by `if (bindingPromise) await bindingPromise` and `bindIdentity`'s `finally` clears the promise. condition is always true at evaluation. dead code in current control flow. harmless. comment intent if guarding against re-entrancy not present in current handler model.

## requested artifacts before merge

1. exact failing invocation: command, OpenCode version/commit, hcom commit.
2. plugin log from failing run showing `HCOM_PROCESS_ID` set, `HCOM_LAUNCHED !== "1"`, no `plugin.session_created`, no `plugin.bound`, no `session_bindings` row.
3. matching log from PR branch with same invocation showing `plugin.bound` and `session_bindings` row.
4. plugin-level regression test driving the asymmetric env state (or documented manual repro path).
5. fix or removal of `plugin.chat_message_unbound` WARN for the non-hcom-launched OpenCode case.

## status

merge not justified by current evidence on `main`. hypothesized real fix scope: a non-standard env-inheritance edge case the description does not specify. independent value: `chat.message` guard tightening (item: retain). other changes pending verification.

mmkzer0:
@aannoo thanks for the repro attempt. I think the important difference is that the failing path was not the same as the smoke test in your comment.

The observed failure happened from Codex via `hcom run partner-setup`. That script is our session-partner bootstrap helper: it looks for an existing co-located OpenCode partner, and if none exists it spawns one, records it under `~/.hcom/session-partners/<owner>-<dirhash>.state`, then later task dispatch sends work to that partner.

The spawn path is `~/dev/code-configs/hcom-scripts/partner-setup.sh:88`:

```sh
hcom opencode --headless --go ${HCOM_NAME:+--name "$HCOM_NAME"} \
  --hcom-system-prompt "$HCOM_SP" -- --agent "$PARTNER_AGENT"
```

Your test used `--hcom-prompt ...`. That matters in current launcher code:

- `--hcom-prompt` becomes `params.initial_prompt`, and for OpenCode launcher injects `--prompt <text>` (`src/launcher.rs:771-799`).
- `--hcom-system-prompt` becomes `params.system_prompt`, but launcher currently only applies that for Gemini (`src/launcher.rs:745-750`). For OpenCode it is effectively ignored.

So the partner-setup path is “OpenCode headless + `--agent` + no OpenCode `--prompt`”, while the smoke test was “OpenCode headless + `--agent` + `--prompt`”. That could explain why your run followed the normal `session.created` path while the partner spawned from Codex got stuck/unregistered.

I agree the PR should not claim this reproduces for every `--agent` launch. More precise claim: this was observed with the `partner-setup.sh` launch shape above. Best next repro is to run that exact command/path from a Codex-launched hcom session and compare `hcom list`/plugin log for whether `opencode-start` runs and whether a `session_bindings` row is created.

mmkzer0:
@aannoo follow-up after more repro work today.

I can no longer reproduce the full failure mode on current local `main` + OpenCode `1.14.31`.

Original failure observed yesterday:

- `hcom run partner-setup` from Codex spawned OpenCode via:
  `hcom opencode --headless --go --hcom-system-prompt ... -- --agent <partner-agent>`
- The spawned partner appeared pty-only / unregistered:
  `session_id: (none)`, `bindings: pty`, `transcript: (none)`
- Terminal was stuck with raw `<hcom>` text, `ready=false`, no usable OC term view, and high CPU.

What I verified today:

- `partner-setup.sh` itself looks sound. `--agent X` and `PARTNER_AGENT=X` both resolve to the same OpenCode launch line.
- With OpenCode `1.14.31`, `--agent` still initially returns a pty-only partner before any OC session exists: `session_id: (none)`, `bindings: pty`, `transcript: (none)`.
- But the first hcom message now usually bootstraps the OC session successfully: `plugin.session_created` -> `opencode-start` -> `plugin.bound`.
- Launching with `--hcom-prompt` also binds cleanly because the initial prompt creates a session immediately.
- I tested both default `m2p7-executor` and non-default `dsv4flash-executor`; both now recover.

So my earlier “messages never deliver” statement was too strong for current OpenCode. The hard failure may have depended on an OpenCode-side prompt/session startup bug that changed with the recent OC `1.14.31` update, or with the local Codex/OpenCode updates today.

Why I still think this hardening is useful:

- `partner-setup` readiness currently means PTY/process readiness, not OC hook/session readiness.
- `--agent` launches can still sit in an observable unbound window until the first prompt/message creates a session.
- This patch makes the plugin binding path less dependent on only the earliest `session.created` timing and hardens startup when the first real session event arrives later.
- Even if current OC recovers, avoiding a pty-only unbound window is safer for hcom partner orchestration and reduces dependence on fragile first-message PTY bootstrap behavior.

Happy to narrow the PR description from “fixes stuck launch” to “hardens `--agent` headless startup binding; original hard failure no longer reproduces on OC `1.14.31`”.

PR #38, opened 2026-05-03
feat: add Kilo, Cline, Antigravity extension, project filter, and stale agent fix

Solar2004:
Adds Kilo Code, Cline CLI, and Antigravity extension support, plus TUI improvements and stale agent fix.


Solar2004:
Starting fresh with clean PR

PR #48, opened 2026-05-09
feat: add Antigravity IDE bridge extension

Solar2004:
## Summary

Add a VS Code extension (hcom-bridge) for Antigravity (Google's VS Code fork) to enable bidirectional messaging between hcom and Antigravity's Jetski agent.

### Changes
- `src/hooks/antigravity.rs` — extension install/verify/remove
- `extensions/hcom-bridge/` — VS Code extension source
- `src/antigravity_extension/` — embedded extension bundle
- Wired into `hcom hooks` management (status/add/remove)
- Architecture docs in `antigravity-integration.md`


aannoo:
### Correctness Blockers

1. **Extension does not import `path`.**

   `extensions/hcom-bridge/src/extension.ts` imports only `vscode`, `fs`, `HcomClient`, `JetskiBridge`. It then calls `path.join(home, '.hcom', 'extensions')` in `logToFile`. `tsc` rejects this under `strict: true`, and `extensions/hcom-bridge/build.sh` runs `tsc` first, so the TS source cannot be rebuilt. The committed minified bundle `src/antigravity_extension/extension.js` references the bare `path` global; on first activation the extension throws `ReferenceError: path is not defined`. TS source and committed JS bundle have drifted.

2. **`send --project` is placed after `--`.**

   In `extensions/hcom-bridge/src/hcomClient.ts`'s `send` method, `args = ['send', '@<target>', '--', message]` is built first, then `args.push('--project', project)` appends the flag *after* the separator. hcom's send grammar treats everything after `--` as message text; `--project foo` becomes part of the message. Project scoping through the bridge's `send` is non-functional. (For reference, `listAgents` in the same file builds `['list', '--names']` first and then pushes `--project` — that ordering is legal; bug is `send`-specific.)

3. **`listen --project` dependency does not exist.**

   The extension's `spawn` method invokes `hcom listen --json --name <name> --project <project>`. PR #49's diff does not modify `src/commands/listen.rs`. A project-configured bridge spawns with an unknown flag, exits non-zero, and the bridge's restart timer puts it into a 3-second spawn loop.

4. **Registration is non-atomic.**

   `hcomClient.ts::startAgent` runs `hcom start --name <name>` first, then `hcom config -i <name> project <project>`. Between the two calls the instance row exists with `project=NULL`. Under PR #49's NULL-as-wildcard semantics, messages from any project can target the just-registered peer during that window. The step-2 promise also swallows errors; if step 2 fails the agent stays cross-project with no retry.

5. **`extensions.json` update is an unlocked read-modify-write.**

   `src/hooks/antigravity.rs` reads `~/.antigravity/extensions/extensions.json`, parses, mutates, serializes, writes. No file lock or atomic-rename. Concurrent installs or IDE writes corrupt the registry. The entry written also embeds `"$mid": 1`, which is a VS Code-internal URI marshalling sentinel — undocumented for external use and subject to change.

6. **Committed minified bundle creates two sources of truth.**

   `src/antigravity_extension/extension.js` is the esbuild output of the TS under `extensions/hcom-bridge/src/` and is embedded via `include_str!` in `src/hooks/antigravity.rs`. Per-change workflow requires editing TS, running `build.sh`, copying the output, and committing both. The two have already drifted (per Blocker 1). `cargo build` should regenerate the blob or fail; tests should diff the embedded blob against a rebuild.

7. **Uses undocumented command APIs despite declaring a typed API.**

   `extensions/hcom-bridge/src/jetskiBridge.ts` calls `vscode.commands.executeCommand('antigravity.startNewConversation')` and `vscode.commands.executeCommand('antigravity.sendPromptToAgentPanel', ...)`. Neither command id is documented in Antigravity's public surface; precedent for instability includes the [agentSessions service-rename incident](https://discuss.ai.google.dev/t/bug-agent-manager-freezes-on-new-chat-unknown-service-agentsessions-leading-to-silent-extension-host-crash/136921). Meanwhile, Antigravity's locally installed `vscode.d.ts` exposes a typed `vscode.antigravityExtensibility.sendToAgentPanel({ message, autoSend, files })` API — and the PR's own `extensions/hcom-bridge/src/antigravity.d.ts` already declares it. The runtime should call the typed API:

   ```typescript
   await vscode.antigravityExtensibility.sendToAgentPanel({
       message: `[hcom message from @${sender}]\n\n${text}`,
       autoSend: true
   });
   ```

   with the reverse-engineered command path only as a versioned fallback.

8. **`EXTENSION_VERSION` hardcoded in three places.**

   A Rust `const` in `src/hooks/antigravity.rs`, plus two `package.json` files. Bumping requires three coordinated edits plus a Rust rebuild. No compatibility detection against Antigravity itself.

PR #49, opened 2026-05-09
feat: project isolation — agents scoped by --project label

Solar2004:
## Summary

Agents with the same project only see each other's messages. Agents with no project label see everyone (backward compatible).

### Changes
- DB schema v18 adds `instances.project` column
- `SenderIdentity.project` in `src/shared/identity.rs`
- `compute_scope()` filters instances by project boundary
- `--project X` flag across: `hcom send`, `hcom list`, `hcom launch`, `hcom events`, `hcom transcript`
- TUI launch panel: project field overlay (`p` key), project filter (`F` key)
- Config: `[launch.project]` in TOML, `HCOM_PROJECT` env var
- Bootstrap PROJECT_NOTICE injected when project is set


aannoo:
thanks but seems to overlap what already exists for project isolation ie HCOM_TAG or HCOM_DIR.

AI review:
## PR #49 - `--project` Isolation

Branch: `feat/project-isolation`

Approximate size: `+649/-173`, 34 files

### Verdict

Delete or rebrand `--project`. Do not keep `delivered_to` as a standalone "good idea" unless hcom intentionally keeps a filtered-broadcast feature or wants durable recipient receipts.

The new `--project` axis overlaps `HCOM_TAG`, falls short of hard isolation, misses several producers/readers, and creates a confusing third scope model beside `HCOM_DIR` and `HCOM_TAG`. `delivered_to` is not independently compelling for normal hcom broadcasts; it is mainly the mechanism PR49 needs because it invents "broadcast, but only to this project."

### What The PR Adds

- `project` column/config plumbing in several local paths.
- `--project` support on some commands.
- `delivered_to` allowlist in event JSON (`src/db/events.rs::should_deliver_to`).
- TUI project filter/launch panel behavior.
- Partial remote launch project propagation.

### What Works

The `delivered_to` idea is useful only for a specific feature shape, not as a general delivery fix. hcom already has per-instance waterline delivery through `instances.last_event_id`; new agents are initialized at the current event id and do not suddenly receive all prior broadcasts.

The actual benefit only appears if hcom adds a filtered-broadcast feature like PR49's `--project`. In current hcom, a normal `scope: "broadcast"` means "everyone", and delivering it to every existing agent is correct.

PR49 tries to create a new kind of broadcast: "broadcast, but only inside project X." It filters the recipient list at send time, but the event is still stored with broad broadcast semantics. A read-side `delivered_to` gate is what makes that filtered broadcast enforceable later: compute the allowed recipients at send time, store them, and have unread reads check that list before delivering the event.

So the useful part is not "late joiners do not get old messages." They already do not. The useful part is narrower: "PR49's proposed project-filtered broadcast has an allowlist that read-side delivery can enforce." If hcom deletes `--project` and does not add filtered broadcasts to tags, then `delivered_to` should probably not be treated as required delivery infrastructure; at most it becomes a receipts/counts/debugging field.

### Blockers

1. **Subscriptions bypass the new gate.**

   `src/db/subscriptions.rs::send_sub_notification` builds its own `delivered_to` from `SELECT name, tag FROM instances` with no project filter. When `events sub --on-hit "<command>"` fires (and `--on-hit` can run `hcom send`), the resulting message reaches every instance in the DB regardless of project. Subscriptions are an active delivery channel, not just a read view — this is a back-door past the gate the rest of the design enforces.

2. **Thread reuse bypasses project gating.**

   `compute_scope` takes `sender_project` and prunes candidate instances, but at `src/commands/send.rs:276`, when the scope is broadcast and a thread is set, the effective recipient set is replaced by `db.get_thread_members(thread)`. `get_thread_members` is not project-filtered. Mixed-project or pre-project threads place cross-project recipients directly into `delivered_to`, which then legitimately pass the read-side gate.

3. **TUI filter can act on the wrong agent.**

   `src/tui/render/agents.rs` builds a filtered list from `app.data.agents` and renders rows from it with `is_cursor = app.ui.cursor == i` where `i` is the filtered index. `cursor_target()` at `src/tui/app.rs:126-129` indexes into the unfiltered `app.data.agents`. With agents `[a (no project), b (dev)]` filtered to dev, the user sees row 0 = b, presses an action, and the handler resolves through `agents[0] = a`. The wrong agent gets killed/messaged/configured.

4. **CLI and TUI disagree on NULL project.**

   CLI paths (`src/messages.rs`, `src/commands/list.rs`) use `.unwrap_or(true)` on the project filter, treating NULL as global/wildcard. TUI filtering uses exact-match (`a.project == proj`), which excludes NULL. Users cannot infer the product rule because the two surfaces ship contradicting interpretations.

5. **Sender-controlled `--project` is not isolation.**

   `src/commands/send.rs:903` resolves `sender_project = args.project.or sender_identity.project()` — the CLI flag overrides the sender row. Any registered project-A agent can pass `--project B` and produce a project-B `delivered_to` allowlist. Fine for a soft selector; fatal if sold as an authorization boundary.

6. **Env/config parity with tag is missing.**

   `HCOM_PROJECT=foo hcom claude` puts `HCOM_PROJECT=foo` in the child's environment via `config::FIELD_TO_ENV` + `to_env_dict`, but `src/commands/launch.rs` reads project only from the `--project` CLI flag and writes `project=NULL` on the instance row. Tag has env→config fallback in the launcher; project does not. Under Blocker 4's NULL-wildcard CLI semantics, the agent is then wildcard-accessible despite being launched with a project env var.

7. **Command coverage is partial — and inconsistent with the PR description.**

   PR body claims `--project` added across "send, list, launch, events, transcript." Verified against the diff: `send`/`list`/`launch` filter correctly. `events` and `transcript` accept the flag but their underlying queries do not filter by project. `listen`, `archive`, `bundle` are not touched. Raw event surfaces and subscriptions are especially important since `--on-hit` is an active delivery channel.

8. **`PROJECT_NOTICE` claimed but missing.**

   PR body says "Bootstrap PROJECT_NOTICE injected when project is set." `src/bootstrap.rs` in this diff contains no `PROJECT_NOTICE` block. Agents launched with a project receive no in-bootstrap statement of what project isolation means.

9. **Relay roster propagation is partial.**

   `src/relay/control.rs` forwards `--project` for new remote launches via `RemoteLaunchRequest::project`. But `src/relay/push.rs::build_push_payload` and `src/relay/pull.rs::import_remote_events` do not include the `project` column in their SELECT/INSERT lists. Existing peer rosters imported across devices arrive with `project=NULL` — wildcard under Blocker 4's CLI rules. Going-forward launches honor project; the existing population leaks.

10. **`hcom start --project` not added.**

    `src/commands/start.rs` only adds `None, // project` placeholders. The Antigravity bridge in PR #48 cannot register identity + project atomically.

11. **Tests are missing and some coverage regresses.**

    The PR lacks behavioral tests for cross-project broadcast, direct mention, NULL wildcard, subscriptions, relay propagation, thread reuse, and TUI filtered actions. It also deletes `send_message_thread_without_members_errors` and the thread seed/reuse-membership test from `commands/send.rs` — thread membership is exactly the surface Blocker 2 depends on, and the right move was to extend those tests with project cases, not delete them.

### Deeper Design Issue

`--project` creates a third scope axis:

- `HCOM_DIR`: hard isolation, different DB/files/hooks.
- `HCOM_TAG`: soft grouping and addressing.
- `--project`: partial SQL label, no addressing, incomplete command coverage.

That is too many concepts for the same user need. The better hcom design is:

1. Keep `HCOM_DIR` for hard separation.
2. Use `HCOM_TAG` for soft grouping/addressing as it exists today.
3. Only add `delivered_to`-style gating if hcom deliberately designs a new tag-scoped broadcast mode.
4. Avoid a separate `--project` label unless there is a proven need for multiple simultaneous tags/groups.

### What To Keep

- The TUI launch panel idea, once filtered cursor/action behavior is correct.
- Some project-label UX may be useful if rebranded honestly as a soft filter, but hcom probably already has this in `tag`.
- `delivered_to` only if hcom deliberately keeps filtered broadcasts or wants explicit recipient receipts. It should not be preserved just because PR49 added it.

### Required Shape

Preferred path:

1. Delete `--project`.
2. Do not add `delivered_to` unless tag-scoped broadcast is explicitly designed.
3. If tag-scoped broadcast is designed, extract one canonical recipient computation function used by direct sends, threads, subscriptions, relay/control messages, and future scoped producers.
4. Add tests for broadcasts, mentions, tags, NULL/global behavior if any, subscriptions, threads, relay, and TUI actions.

Alternative path if project must stay:

1. Rebrand as soft grouping, not isolation. Reconcile the PR description with the actual command coverage; drop claims about `events`/`transcript` filtering and `PROJECT_NOTICE` unless those are added.
2. Document NULL-as-wildcard clearly in product help and bootstrap text. Pick one rule for NULL and apply it in both CLI and TUI.
3. Fix subscriptions, thread reuse, relay push/pull, env/config parity, TUI cursor mapping, and command coverage.
4. Or commit to strong isolation by removing sender override, carrying project in event envelopes, gating all raw event surfaces, and eliminating NULL wildcard by default.

### Merge Recommendation

Do not merge as "isolation." Reject `--project` as a third scope axis. Keep `delivered_to` only if the accepted replacement design explicitly includes scoped broadcasts or durable recipient receipts.

PR #53, opened 2026-05-21
Fix Codex hook verify with empty-matcher peer groups

mmkzer0:
# Fix Codex hook verify with empty-matcher peer groups

## Summary

This fixes a Codex hook setup false failure that showed up as `hcom codex`
failing post-write verification with:

```text
hcom hook command not found
```

Found it while tracing a live `hcom codex` launch failure after hook setup
had already written hcom's Codex groups into `hooks.json`.

## Cause

The regression was exposed when context-mode started writing
`"matcher": ""` groups for `UserPromptSubmit` and `Stop`.

hcom writes those events into groups with no `matcher` key. The verify path had
historically treated an absent matcher and an empty-string matcher as
equivalent, then used the first matching group it found. Once a context-mode
empty-matcher group appeared before hcom's no-key group, verify checked the
wrong group and failed even though hcom's hook had been written.

The same investigation also found stale pre-Codex-0.129 hcom hook entries in
the old `{"type":"cmd","cmd":...}` format. They had been left behind silently
since hcom moved to the current `command`-keyed format.

## Fix

- make Codex hook verification use the same matcher grouping rule as the write
  path for no-matcher events
- clean stale legacy hcom `cmd`-format entries when the current Codex hook
  format is supported
- add regression coverage for both the context-mode empty-matcher case and the
  legacy cleanup path

## Validation

Automated:

```text
cargo test test_context_mode_empty_matcher_does_not_block_verify -- --nocapture
cargo test test_legacy_cmd_hooks_cleaned_on_new_codex -- --nocapture
cargo test hooks::codex::tests -- --nocapture
```

Manual:

- reproduced from the live Codex hook state that broke setup
- verified that `hcom codex` launches successfully again after the fix

### DIFF
--- a/src/hooks/codex.rs
+++ b/src/hooks/codex.rs
@@ -679,7 +679,7 @@ fn is_legacy_hcom_codex_cmd_entry(hook: &Value) -> bool {
         })
 }
 
-/// Remove legacy `"cmd"`-keyed hcom hook entries, leaving user-owned hooks intact.
+/// Remove recognized legacy `"cmd"`-keyed hcom hook entries.
 /// Only called when Codex >= CODEX_HOOKS_FEATURE_RENAME_VERSION, which is when
 /// the current `"command"`-keyed format is known to be supported.
 fn remove_legacy_hcom_cmd_hooks_from_json(existing: &mut Value) {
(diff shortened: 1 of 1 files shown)


aannoo:
Thanks

PR #61, opened 2026-06-03
Preserve OpenCode model variants for hcom messages

### LINKED ISSUE #60: OpenCode hcom messages drop model variants and reset effort
OpenCode stores reasoning effort as the model `variant` - for example `variant=high` or `variant=xhigh`. Normal user turns keep that value, but hcom-injected OpenCode turns are currently sent through `promptAsync` with only `agent`, `model`, and `parts`.

This is OpenCode-specific: OpenCode carries reasoning effort as the prompt's model `variant`, and hcom delivers messages via `promptAsync` where that `PromptInput.variant` is dropped. Other harnesses pass effort through CLI/session mechanisms, so they do not hit this path.

Because `src/opencode_plugin/hcom.ts` tracks `currentModel` as only `{ providerID, modelID }`, the injected user message is saved without a variant. In my OpenCode sessions, those blank-variant hcom turns reset the session effort away from the user's current setting, so agents silently downgrade reasoning effort mid-session.

Environment checked:

- hcom: local build from `0.7.19`; upstream checked at `origin/main` `3db0398` / `v0.7.20`
- OpenCode: `1.15.12-202605282246-85ee30`
- OS: Linux / WSL2

Expected: hcom-injected OpenCode turns keep the active model variant/effort.

Actual: hcom-injected OpenCode turns are saved with a blank variant, and the session effort resets to the default.

The code path looks sufficient to explain the bug: `src/opencode_plugin/hcom.ts` receives `input.variant` in `chat.message`, but drops it when normalizing `currentModel`, then calls `promptAsync` without a top-level `variant`. I also confirmed locally that normal user messages had `variant=high`, while recent `<hcom>` messages in the same session had a blank variant.

I opened [PR #61](https://github.com/aannoo/hcom/pull/61) with the proposed fix:

- add `variant?: string` to `PromptModel`
- preserve `input.variant` in `chat.message` via `normalizePromptModel(input.model, input.variant)`
- pass `variant: currentModel?.variant ?? undefined` as the top-level `promptAsync` variant
- keep the `model` payload sent to OpenCode limited to `providerID` and `modelID`
- parse `--variant` in the CLI model fallback path

Manual verification with a rebuilt hcom showed hcom-delivered OpenCode turns persisted with the active variant, for example `variant=high`, instead of blanking the field.

azkore:
OpenCode carries reasoning effort as the prompt model `variant` (`high`, `xhigh`, etc.). hcom's OpenCode plugin was tracking only `providerID` and `modelID`, so hcom-delivered `promptAsync` turns were saved without a variant. In active sessions this can reset the effort back to the model default and silently lower reasoning quality.

This keeps the current OpenCode variant with the tracked model, preserves `input.variant` from `chat.message`, and passes that variant back as the top-level `promptAsync` variant when hcom delivers queued messages. The `model` payload sent to OpenCode stays limited to `providerID` and `modelID`, and `--variant` is read from launch args as a fallback for sessions started with an explicit variant.

I also verified the fix manually against a live OpenCode session: before the fix, `<hcom>`-delivered turns were saved with a blank variant; after the fix, they persist the active variant, for example `variant=high`, in the OpenCode DB and the session effort no longer resets. I ran `cargo test opencode`; 55 matching tests passed.

Fixes #60.

### DIFF
--- a/src/opencode_plugin/hcom.ts
+++ b/src/opencode_plugin/hcom.ts
@@ -9,6 +9,7 @@ const LOG_PATH = `${HCOM_DIR}/.tmp/logs/hcom.log`
 type PromptModel = {
   providerID: string
   modelID: string
+  variant?: string
 }
 
 // Best-effort fallback for non-hcom/manual plugin runs.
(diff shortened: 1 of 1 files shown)


aannoo:
thanks

PR #65, opened 2026-06-09
Fix delayed hcom message delivery

azkore:
Subscription notifications could be logged as `[hcom-events]` messages without waking the subscribed agent. That left agents waiting until some unrelated status change or message wake made them read the queued notification. OpenCode had a second race where a wake that arrived during an active delivery read or while a prompt injection was waiting for transform ack was ignored.

This fixes both paths. Subscription notifications now wake only the matched subscriber after the message is logged, and OpenCode queues one pending delivery wake and drains it after the current read, after deferred ack, or after `promptAsync` rejects.

The Rust path adds a TCP-probe test proving only the subscribed target is woken; the OpenCode replay path has no unit test because it depends on plugin runtime sequencing and was validated against a state-machine repro of the in-flight, pending-ack, and rejection sequence. Checked with `cargo fmt --check`, `cargo test db::subscriptions::tests::test_send_sub_notification_wakes_target_instance`, and `bun build src/opencode_plugin/hcom.ts --external @opencode-ai/plugin --external @opencode-ai/sdk --outfile /tmp/opencode/hcom-plugin.js`.

### DIFF
--- a/src/db/subscriptions.rs
+++ b/src/db/subscriptions.rs
@@ -1008,7 +1008,17 @@ fn send_sub_notification(db: &HcomDb, caller: &str, message: &str) -> bool {
     };
 
     let text = format!("@{} {}", full_name, message);
-    send_system_message(db, "[hcom-events]", &text).is_ok()
+    let Ok(delivered_to) = send_system_message(db, "[hcom-events]", &text) else {
+        return false;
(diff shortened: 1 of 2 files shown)


aannoo:
thanks

Issue #11, opened 2026-03-13
TUI launch broken for single agent


monofunc:
Launching a single agent from TUI always runs it headless regardless of settings. In Ghostty, no window or tab opens but the agent shows as running, can't receive messages and won't terminate. In Terminal.app it shows `launch_failed` instead. Launching two or more agents works correctly.

PR #68, opened 2026-06-28
feat(omp): add full support for oh-my-pi

mmkzer0:
This was rather straightforward given that Pi was supported already,
so general changes were small and only omp specifics (omp hook and plugin)
make up a considerable amount of new code.

## Summary

Adds full native support for omp (Oh My Pi) to hcom.
Mirrors the existing Pi integration while isolating its config, hooks and transcripts.

## omp Specifics

- Registered omp with a dedicated launch.omp config block and status mapping.
- Added src/hooks/omp.rs and src/omp_plugin/hcom.ts. Installs the TS plugin to ~/.omp/agent/extensions/hcom.ts and routes omp-* lifecycle
 events.
- Uses the PiJsonl backend format but distinct OmpSessions roots for discovery (defaulting to ~/.omp/agent/sessions).
- path res -> Reuses PI_CODING_AGENT_DIR and PI_CODING_AGENT_SESSION_DIR, correctly reflecting OMP's upstream inheritance of Pi's environment variables.
- Added .omp to PROTECTED_HCOM_DIR_COMPONENTS to shield the hcom database from AI sandbox policies.

## General Changes

- Delivery: Updated src/delivery.rs to include Tool::Omp in the plugin-delivery model (PTY bootstrap for the first message, then plugin takeover) and compact wake injection.
- TUI/RPC: Mapped OMP into the canonical Tool enum and launch cycle, ordered adjacent to Pi.
- Tests: Refactored the hook_names_are_disjoint test to iterate integration_spec::ALL dynamically, replacing a brittle hardcoded array.

## Testing

- cargo fmt --all -- --check
- cargo clippy -- -D warnings
- cargo test (1844 tests passing)
- omp <-> omp and omp <-> OpenCode live system tests passed

## Issues encountered

Parallel Test Leakage: Found that manual std::env::set_var/remove_var calls in
Pi/OMP transcript tests were vulnerable to race conditions since Rust tests run concurrently.
Fixed by extending the RAII EnvGuard to track PI_CODING_AGENT_DIR and PI_CODING_AGENT_SESSION_DIR, 
guaranteeing panic-safe isolation.

## omp <-> omp demo

demo


aannoo:
Nice! Here is AI review:

PR #68 Review

## Root cause

Most findings trace to one thing: the plugin (`src/omp_plugin/hcom.ts`) and the
integration spec were copied from the Pi integration without adapting to
oh-my-pi's diverged extension API and session lifecycle. Confirmed against
`@oh-my-pi/pi-coding-agent` vs `@earendil-works/pi-coding-agent`:

- input-handler result: omp uses `{ handled?, text?, images? }`; pi used `{ action }`.
- `InputEvent` has no `streamingBehavior` in omp.
- `SessionShutdownEvent` has no `reason` in omp (pi's did).
- omp delivers in-process session changes via a dedicated `session_switch`
  event; pi folded them into `session_shutdown` (with a reason) + `session_start`.

#1, #3, #4, #11 below are direct consequences of these.

## P1 — Blocking

1. **Use omp's input-hook result shape**
   omp expects `{ handled, text, images }`, not `{ action: ... }`. Return `{}`
   to continue, `{ text }` to transform, `{ handled: true }` to swallow.
   `{ action: "handled" }` does not set `handled`, so a bodyless `<hcom>` with no
   pending falls through as a literal prompt to the agent. Also remove the
   `event.streamingBehavior` access (input handler) — the field doesn't exist on
   omp's `InputEvent`.

2. **Add omp to the automatic-delivery bootstrap**
   The auto-delivery branch in `src/bootstrap.rs` (~L466) lists pi but not omp,
   so omp falls through to `DELIVERY_ADHOC` ("Messages do NOT arrive
   automatically") — contradicting the native plugin + PTY delivery. The agent is
   told to manually `hcom listen`. Add `omp` to the launched-tool list.

3. **Reset and rebind on `session_switch`**
   The plugin binds on `session_start` and resets on `session_shutdown`, but omp
   emits a separate `session_switch` (reason `new`/`resume`/`fork`) for in-process
   switches. Without handling it, delivery and status stay attached to the
   previous session. Reset session state and rebind on `session_switch`.

4. **ACK the bodyless PTY-wake transform path**
   Transforming `<hcom>` into pending messages sets `pendingAckId`, but nothing
   acks it: the transform path goes through neither the extension-source input
   ack nor `deliverPending`'s ack. `deliverPending` then early-returns forever on
   `pendingAckId !== null`, so later delivery stays blocked. ACK from an event
   tied to the transformed submission, such as `before_agent_start`.

5. **Normalize all omp resume aliases**
   omp's resume aliases are `--resume`, `-r`, `--session`, `--continue`, `-c`.
   `merge_omp_args()` leaks `-r`, `-c`, and `--resume=<id>` (only the
   `--session=`/`--session-id=`/`--session-dir=`/`--fork=` equals-forms are
   stripped), and can discard an unrelated following flag after a valueless
   `--session`. Strip all aliases + equals-forms, consuming an optional value only
   when the next token is value-like. (Also: the doc comment claiming omp "lacks
   `--fork` and `--session`" is wrong — both exist in omp's flag tables.)

6. **Reject modes that bypass the interactive extension**
   `OMP_REJECTED_ARGS` only rejects `-p`/`--print`. omp also accepts `--mode acp`,
   `--mode rpc`, and `--no-extensions`, which either skip `hcom.ts` or wait on a
   TUI readiness marker that never appears. Reject them.

7. **Include XDG + profile session roots in discovery**
   omp flattens `~/.omp/agent/sessions` to `$XDG_DATA_HOME/omp/sessions` when XDG
   is set, and named profiles relocate the agent dir. Discovery
   (`disk_search_roots` / resume / adopt) only checks
   `PI_CODING_AGENT_SESSION_DIR`, `PI_CODING_AGENT_DIR/sessions`, and
   `~/.omp/agent/sessions`. Add the XDG and named-profile roots.

## P2 — Should fix

8. **Do not let Pi claim omp sessions from the shared session-dir override**
   `find_session_on_disk` probes Pi (step 8) before omp (step 9), and both read
   `PI_CODING_AGENT_SESSION_DIR`. When that override points at an omp dir, an omp
   session resolves as `tool="pi"`, and transcript search drops shared-root files
   as ambiguous. Preserve provenance, inspect a reliable marker, or require a tool
   qualifier. Low frequency — only when the shared override is set.

9. **Resolve the plugin install dir from the final child env**
   Installation resolves `~/.omp/agent/extensions` (or
   `PI_CODING_AGENT_DIR/extensions`) before the child environment — profile,
   project-local isolation, `~/.hcom/env`, launch overrides — is finalized. With a
   profile/XDG/isolation, omp loads extensions from a different directory (e.g.
   the active profile's agent dir). Resolve from the final child env or pass
   `--extension <absolute hcom.ts>`. Default no-profile case is unaffected.

10. **Enable omp fork support**
    omp supports a value-taking `--fork`, but the integration declares
    `fork: None`. Add the fork mapping and resume/fork merge tests.

mmkzer0:
Follow-up after PR review
- aligned OMP plugin input results with the Oh My Pi SDK
- include OMP in bootstrap auto-delivery paths
- handle OMP session switches and plugin delivery ACKs
- inject the OMP extension explicitly at launch
- fix OMP resume argument merging
- discover OMP transcripts from PI_CODING_AGENT_* roots
- protect OMP hook install/remove from clobbering user files
- add regression tests for the affected paths

Only one correction on fork support: OMP does not expose --fork; I kept fork: None

aannoo:
Made a few extra commits:

## OMP integration

- Adds Oh My Pi to package metadata, marketplace manifests, README documentation, CLI help, configuration help, and agent-messaging skill documentation.
- Documents `hcom omp`, `omp_args`, automatic message delivery, and OMP session forking.
- Rejects `omp acp` because ACP mode does not provide the interactive TUI required by hcom.

## Launch readiness

- Replaces OMP’s unreliable screen-text readiness check with extension-bind readiness.
- Adds a database query for notify endpoints by kind.
- Marks OMP ready only after its plugin notify endpoint is registered.
- Wakes the PTY delivery loop when the plugin binds.
- Adds a longer grace period while waiting for plugin binding.

## Message delivery

- Queues wakes received while delivery or acknowledgement is already in progress.
- Replays queued delivery after the active operation finishes.
- Keeps delivery gated until acknowledgement succeeds.
- Retries failed acknowledgements during reconciliation.
- Acknowledges bodyless `<hcom>` input transformations from `before_agent_start`.
- Rebinds the plugin after OMP creates a branched session.

## Resume and fork

- Adds native OMP forking through `omp --fork <session-id>`.
- Removes prior resume/fork controls before applying the requested operation.
- Stops persisting hcom’s injected OMP extension path in launch arguments.
- Removes stale managed extension arguments during replay and injects the current extension path at launch time.

## Pi and OMP session discovery

- Separates Pi-exclusive, OMP-exclusive, and shared session roots.
- Keeps `PI_CODING_AGENT_SESSION_DIR` exclusive to Pi.
- Supports OMP default and named-profile session trees.
- Supports initialized OMP XDG data directories.
- Supports `PI_CONFIG_DIR`, `OMP_PROFILE`, legacy `PI_PROFILE`, and `PI_CODING_AGENT_DIR`.
- Attributes shared-root sessions using `.pi` or `.omp` path markers.
- Reports markerless shared-root sessions as ambiguous instead of guessing.
- Uses the same root resolution for transcript search and session adoption.
- Treats Android/Termux as XDG-capable to match OMP’s Bun runtime behavior.

## Lifecycle

- Extends duplicate unchanged status-event suppression from Pi to both Pi and OMP.

## Tests

- Adds coverage for plugin-bind launch readiness and notify wake-up.
- Adds OMP delivery acknowledgement, queued-wake, and branch-rebinding checks.
- Adds OMP resume, fork, and managed-extension argument tests.
- Adds Pi/OMP root ownership, profile, XDG, shared-root, and ambiguity tests.
- Updates OMP PTY readiness handling to use plugin binding.
- Extends test environment restoration for OMP and Pi profile/config variables.

PR #81, opened 2026-07-11
feat(grok): add first-class Grok Build integration

kiala9:
## Summary

Adds first-class **Grok Build (`grok`)** support to hcom: hook install, PTY launch/resume/fork, message delivery, transcript parsing, config, docs, and focused tests.

Validated on WSL/Windows Terminal against Grok Build (hooks + PTY + multi-agent send/reply).

## Delivery protocol (aligned with Grok StopHookJson)

Grok discards stdout on observe-only hooks (SessionStart / UserPromptSubmit / PostToolUse). The only reliable model-injection gate is **Stop → `hookSpecificOutput.additionalContext`**.

```text
hcom send → notify → PTY injects short wake sentinel only ("hcom: wake") + Enter
                   → turn starts (pending kept; no force-ack)
                   → genuine Stop(end_turn) → additionalContext = full <hcom> body + commit_delivery_ack
                   → agent replies via hcom send
```

Key rules after owner review alignment (`d482ea5`):

1. **PTY = wake only** — never blind full-body inject / Ctrl-U / force-ack
2. **Observe hooks** — no delivery on UPS/PostToolUse/SessionStart stdout
3. **Stop only** — `additionalContext` (+ snake_case twin); never `followup_message`
4. **Session-end Stop** — skip deliver/ack (`channel_closed`, shutdown, …)
5. **`$GROK_HOME`** for hooks path; dual UPS races avoided by keeping pending until Stop

## What works

- `hcom grok` / `hcom grok-build` — launch with hooks install
- `hcom hooks add grok` / `hcom hooks remove grok` — `~/.grok/hooks/hcom.json`
- `hcom r <name>` / `hcom f <name>` — `--resume` + `--fork-session`
- Message delivery to/from other tools (Claude, Codex, Cursor, …)
- `hcom transcript` via `~/.grok/sessions/**/updates.jsonl`
- Wake-mode delivery + Enter retry on slow WSL (no premature ST_ACTIVE confirm)

## New files

- `src/hooks/grok.rs` — hook install/verify/remove + lifecycle (Stop gate)
- `src/transcript/grok.rs` — `updates.jsonl` parser

### Automated

- [x] `cargo test grok` — hooks, Stop additionalContext shape, host detection, delivery wake tests
- [x] `cargo fmt` / clippy cleanups on the branch

### Manual (author, home/`~` workspace)

- [x] Wake sentinel only in transcript user turn (`hcom: wake`)
- [x] Synthetic `hcom grok-stop` with pending → stdout `hookSpecificOutput.additionalContext` + log `grok.stop.additional_context` + cursor advance
- [x] Live multi-agent: bozo → reviewfix-mika → `PONG-PROTO` ack (agent also may self-pull via `hcom listen`, which advances cursor; pure mid-turn Stop body inject depends on ending turn without listen)

## Usage

```bash
hcom hooks add grok
hcom grok
hcom send @<name> -- "hello from another agent"
```

## Known limits / follow-ups

- Composer scrape still returns `None` for Grok (wake uses force-Enter after sentinel; no Ctrl-U)
- Full composer ownership / dual-queue with Grok Build submit queue still open
- SessionStart bootstrap must use launch-time channels (`--rules` / skill), not hook stdout
- Agents that `hcom listen` mid-turn advance `last_event_id` and can skip the Stop additionalContext path (listen is intentional CLI consumption)

## Offline patch kit

`~/workspaces/hcom-patches/grok-build-integration` tracks tip `d482ea5` for brew/offline apply.


aannoo:
thanks for the pr looks good. here is AI review may or may not be relevant you can check what you want to do.

# PR 81 review: Grok Build integration

**Verdict: do not merge.** The hook delivery path is structurally broken, the PTY path bypasses every safety mechanism hcom has, and content is mutated with no working fallback. The integration scaffolding (registration, launch, lifecycle, resume, transcripts) has value and can merge after targeted fixes, but the delivery architecture needs a redesign before `released: true`.

## What was reviewed

PR 81 at `0d83923` (27 files, +2133 lines) applied against hcom `main`. Cross-referenced against the Grok Build source at `c5c4ce03` (v0.2.106). No build or test execution — static source analysis only.

## The hook delivery path is broken at the protocol level

### Observe-only events cannot deliver messages

Grok classifies hook events by gate type. `SessionStart`, `UserPromptSubmit`, `PostToolUse`, `SessionEnd`, and notifications are `GateKind::Observe`. The command runner's Observe branch (`xai-grok-hooks/src/runner/command.rs:230-238`) returns `Success` or `Failed` based solely on exit code. **Stdout is never parsed.** It is read, measured for logging, and discarded.

PR 81 emits message payloads from three observe-only events:

- `handle_sessionstart` (`src/hooks/grok.rs:388-434`): emits `additional_context` and `hookSpecificOutput.additionalContext` for bootstrap
- `handle_userpromptsubmit` (`src/hooks/grok.rs:437-482`): emits `followup_message`
- `handle_posttooluse` (`src/hooks/grok.rs:499-519`): emits `additional_context` and `hookSpecificOutput.additionalContext`

None of these outputs reach the model. Every message that attempts delivery through these hooks is silently lost.

### Stop uses the wrong output field

`Stop` is a real blocking gate (`GateKind::Stop`). Its stdout is parsed as `StopHookJson` (`xai-grok-hooks/src/runner/mod.rs:64-75`), which accepts:

```rust
struct StopHookJson {
    decision: Option<String>,
    reason: Option<String>,
    continue_: Option<bool>,       // serde rename "continue"
    stop_reason: Option<String>,   // serde rename "stopReason"
    hook_specific_output: Option<StopHookSpecificOutputJson>,
}

struct StopHookSpecificOutputJson {
    additional_context: Option<String>,
}
```

PR 81's `handle_stop` (`src/hooks/grok.rs:548-551`) emits:

```json
{"followup_message": "..."}
```

`followup_message` is not a field on `StopHookJson`. Serde deserializes this as an empty `StopHookJson` with all `None` fields. The message never reaches the model.

### Acknowledgment fires on pipe flush, not model delivery

`dispatch_grok_hook` (`src/hooks/grok.rs:612-618`) commits the delivery ack after `serde_json::to_writer` + `stdout.flush()` succeeds. This proves only that bytes were written to a pipe. It does not prove Grok parsed them, accepted them, or exposed them to the model.

For observe-only events this is a guaranteed silent loss: hcom marks the message as delivered while Grok drops the output.

For Stop, the wrong field means the same outcome: empty parse, no continuation, message lost, cursor advanced.

### Session-end Stop creates a deterministic loss case

Grok fires a separate session-end Stop when the user exits. `dispatch_session_end_stop` (`xai-grok-shell/src/session/acp_session_impl/stop_gate.rs:112-141`) runs hooks in Stop mode but calls `demote_ignored_blocks` on the results — block decisions are discarded because no turn remains.

PR 81's `handle_stop` does not filter on the Stop `reason`. It checks a `status` field for cancellation strings but does not distinguish `reason: "end_turn"` from `reason: "channel_closed"` / `"shutdown"`. A message pending at session exit will be emitted through `followup_message`, Grok will ignore the output, and hcom will advance the cursor.

## The PTY full-body path bypasses five safety mechanisms

### 1. Prompt readiness disabled

```rust
require_ready_prompt: false,
```

hcom's ready-prompt check ensures the tool's normal input interface is visible. Disabling it means hcom will inject text into any screen state.

### 2. Empty prompt check disabled

```rust
require_prompt_empty: false,
```

hcom's empty-prompt check prevents submitting content when the user has an unsent draft. The PR acknowledges that Grok's input box cannot be scraped — but the correct response is "don't inject," not "inject anyway."

### 3. Ctrl-U without observation

`clear_composer_best_effort` sends Ctrl-U before every Grok delivery. Grok's textarea implements Ctrl-U as delete-from-cursor-to-line-start, not clear-all. Text after the cursor survives. hcom cannot observe the result because it has no composer parser. A user draft can be partially deleted and then combined with the hcom message.

### 4. Force Enter without ownership

After injecting text and sleeping for a length-scaled settle time, hcom presses Enter without verifying that the composer contains exactly the injected text. This is the fundamental ownership check that hcom added and hardened across multiple tools and PRs. Bypassing it means:

- An unrelated user prompt can be submitted
- A partial or corrupted inject can be submitted
- An empty composer can receive a spurious Enter

### 5. Force-ack after failed verification

After three Enter attempts without confirmation, hcom force-acknowledges the pending batch:

```rust
// Last resort: clear bus (composer may still need manual Enter).
if let Some(prepared) = prepare_pending_messages(db, &current_name) {
    commit_delivery_ack(db, &prepared.ack);
}
```

This produces at-most-once delivery under uncertainty: when hcom cannot prove delivery, it discards the messages. This conflicts with hcom's own direction of travel — bounded retries that park delivery while leaving messages pending.

## Content is mutated and truncated

`sanitize_grok_inject_text` (`src/delivery.rs`):

- `<` → `[`, `>` → `]` (breaks XML, HTML, Rust/C++ generics, shell redirection, comparisons)
- Newlines and control characters → spaces (destroys code blocks, stack traces, structured data)
- All whitespace collapsed
- Capped at 2,000 characters

The code comment says "full body still arrives via UserPromptSubmit hooks when present." But UserPromptSubmit is observe-only — stdout is ignored. There is no working fallback for the truncated or mutated content.

## Dual hook ownership creates race conditions

Grok loads Claude-compat hooks from `~/.claude/settings.json` alongside native hooks from `~/.grok/hooks/`. PR 81 adds Grok-specific delivery behavior to both:

- **Native Grok hooks** (`src/hooks/grok.rs`): PostToolUse emits `additional_context`, Stop emits `followup_message`
- **Claude-compat hooks** (`src/hooks/claude.rs`): PostToolUse emits `followup_message` when `is_grok_host()`, Stop delegates to `handle_poll_for_grok` which emits `followup_message`, UserPromptSubmit has Grok-specific body-matching logic

Both hook sets fire for each event. The code tries to coordinate with `prompt_already_carries_hcom_body`, which does 48-char prefix matching and pattern matching for `"request #"` + arrows. This is a heuristic that can false-positive on user content and false-negative when formatting differs between the two representations.

One owner per tool, not two racing owners with heuristic deduplication.

## Other defects

### `$GROK_HOME` ignored

Grok resolves its home as `$GROK_HOME` or `~/.grok` (`xai-grok-config/src/paths.rs:34-38`). PR 81 hardcodes `tool_config_root().join(".grok")` for hooks and `home/.grok/sessions` for transcripts, with `config_dir_env: None` in the integration spec.

### Resume argument handling drops valid values

`merge_grok_args` (`src/commands/resume.rs`) maintains a `VALUE_FLAGS` list but is missing:

- `--worktree [NAME]`
- `--worktree-ref` / `--ref REF`
- `--leader-socket PATH`
- `--debug-file FILE`
- `--system-prompt-override PROMPT`

All non-flag positional tokens are dropped as "stale task prompts." `--worktree feature --ref main` becomes `--worktree --ref`.

### Transcript chunk assembly

`content_text` (`src/transcript/grok.rs:32-55`) `.trim()`s every chunk before use. Streamed chunks like `"Hello "` + `"world"` become `"Hello"` + `"world"` → `"Helloworld"` when joined. Grok's own ACP stream concatenates raw chunks without trimming.

All tools are marked `is_error: false`. Tool results, failure updates, and `_detailed` mode are ignored.

### One-shot mode treated as persistent

`-p` / `--single` is documented as one-shot scripting mode. The PR advertises it as `hcom grok -p "task"` while treating Grok as a PTY-managed persistent agent. A one-shot process exits and cannot receive later messages.

## Delivery choices exposed by the reviewed Grok code

### Two-stage sentinel delivery

Cursor and Copilot use the same basic delivery architecture:

```text
message arrives while the agent is idle
→ hcom proves the normal composer is visible and empty
→ PTY submits only <hcom>
→ bootstrap says <hcom> is not a task: do not answer, do not run tools, end immediately
→ the wake turn reaches its end-of-turn hook
→ the hook supplies the real hcom body
→ the model handles the body in the second stage
```

The host-specific difference is only how the second stage is created:

- **Cursor:** Stop returns `followup_message`; Cursor submits that body as a new user prompt.
- **Copilot:** AgentStop returns `decision:block` with the body in `reason`; Copilot prevents termination and continues with that feedback.
- **Grok:** the reviewed Stop schema accepts `hookSpecificOutput.additionalContext`; static analysis shows that context is fed back into the model at Stop. Grok does not expose Cursor's `followup_message` field, and its exact continuation fields are Grok-specific rather than Copilot's schema.

For Grok, the corresponding idle flow is therefore:

```text
prove Grok's composer is visible and empty
→ submit a small <hcom> sentinel
→ bootstrap makes Grok end the wake turn
→ genuine Stop(reason="end_turn") fires
→ Stop additionalContext contains the real batch
→ Grok processes the batch in the continuation stage
```

For a message that arrives while Grok is already working, the PTY wake is unnecessary:

```text
message remains pending
→ current Grok turn reaches genuine Stop(reason="end_turn")
→ Stop additionalContext supplies the real batch
→ Grok continues with it
```

`UserPromptSubmit` and `PostToolUse` cannot carry the body because their stdout is ignored. Stop-added context is limited to roughly 10,000 characters and eight continuations, and the same boundary is not available on interruption, refusal, maximum-turn termination, API error, or shutdown.

### Historical-Codex-style retrieval

```text
short PTY sentinel or Stop feedback
→ bootstrap/injected context tells Grok to run an hcom read command
→ exact message body arrives as tool output
→ the retrieved batch is acknowledged
```

The sentinel still creates the wake turn, but the second stage retrieves the body through a tool call instead of carrying it directly in Stop output. This avoids the Stop context-size limit and requires separate read and acknowledgment semantics.

### Full-text PTY prompt

```text
Grok is idle
→ prove the normal composer is visible and empty
→ paste the exact message body
→ verify the composer contains exactly that body
→ submit it as a normal user prompt
```

This skips the sentinel/Stop second stage and places the real body directly in the composer. It is idle-only and cannot inject a new prompt while Grok is already processing a turn.

PR 81's current path mutates and truncates the body, cannot prove an empty composer, uses Ctrl-U without observing the result, submits without exact ownership, and force-acknowledges on uncertainty.

### ACP

```text
hcom opens or controls a Grok ACP session
→ messages are sent through the structured session protocol
→ responses and lifecycle events return through ACP
```

This bypasses both PTY sentinel delivery and file-hook payload delivery. Static analysis shows an ACP session implementation but does not establish attachment to an already-running interactive Grok TUI session.

### Other hcom delivery models

| Model | What happens |
|---|---|
| Prompt-start hook payload | Claude PTY, Codex, Kimi, and Gemini attach the real body during the same turn started by the PTY trigger |
| Blocking Stop listener | Claude print/vanilla Stop waits for a message and then resumes Claude with the body; no PTY wake |
| Plugin/extension API | OpenCode, Kilo, Pi, and OMP call a host prompt/message API directly |
| Manual retrieval | The agent runs `hcom listen` without an automatic wake path |

## Merge recommendation

### Merge after fixes

| Component | Fix needed |
|-----------|-----------|
| `Tool::Grok` registration, aliases, config, help | None |
| Hook installation/removal | Resolve `$GROK_HOME` consistently |
| Launch and lifecycle hooks (SessionStart, SessionEnd, PreToolUse) | Remove delivery/bootstrap from SessionStart stdout |
| Status tracking and TUI integration | None |
| Resume/fork | Fix arg handling: preserve unknown flags verbatim, reject `-p`/`--single` |
| Transcript parser | Remove chunk trimming; add real fixtures with chunk boundaries |

### Remove or replace

| Component | Reason |
|-----------|--------|
| PostToolUse message delivery | Observe-only; stdout ignored |
| UserPromptSubmit message delivery | Observe-only; stdout ignored |
| `followup_message` from Stop | Not in `StopHookJson` schema |
| Current full-body PTY implementation | Content mutation, no ownership, no verification |
| `sanitize_grok_inject_text` | Semantic corruption with no working fallback |
| Ctrl-U / force-Enter / double-Enter | Unsafe without composer parser |
| Force-ack after failed delivery | Violates pending-message invariant |
| Grok-specific delivery logic in `claude.rs` | One delivery owner per tool |
| `prompt_already_carries_hcom_body` | Heuristic deduplication for a race that shouldn't exist |
| Current `GrokDeliveryMode` / `HCOM_GROK_DELIVERY` paths | They select delivery paths that are not protocol-correct as implemented |

### Delivery choices requiring implementation work

| Choice | Missing implementation |
|---|---|
| Two-stage sentinel delivery | Add a Grok composer parser and safe `<hcom>` submission for idle wake; deliver the real body only from genuine `Stop(reason=="end_turn")` using Grok's supported continuation schema |
| Active Stop delivery | Leave messages pending until genuine Stop, obey Stop context/continuation limits, and acknowledge only after the supported handoff |
| Historical-Codex-style retrieval | Add a non-destructive read plus exact batch acknowledgment after the tool result is observed |
| Full-text PTY prompt | Add lossless paste, exact composer parsing, post-paste ownership verification, submission correlation, and failure-safe acknowledgment |
| ACP | Define whether hcom creates a separate managed Grok session or can attach to an existing interactive session |
| Bootstrap | Use `--rules`, `--append-system-prompt`, or another supported launch-time channel rather than SessionStart stdout |

kiala9:
@aannoo quick ping on PR #81. Your Jul 21 review was against `0d83923`; I pushed `d482ea5` specifically to address that protocol review.

Main delta:
- observe-only hooks no longer deliver/ack messages
- Grok delivery is Stop-only via `hookSpecificOutput.additionalContext`, no `followup_message`
- session-end Stop skips deliver/ack
- PTY path is wake sentinel only, no full-body paste, Ctrl-U, sanitize, or force-ack
- Claude-compat hooks no longer own Grok delivery
- `$GROK_HOME`, resume flags, transcript chunk trimming, and `is_error` were fixed

Latest head is `d502eb7`; CI #423 is green including Windows build. The remaining explicit tradeoff is that Grok composer scrape is still unavailable, so idle wake uses a short sentinel rather than verified composer ownership. Could you re-review just the post-`0d83923` delta and say whether `released: true` is acceptable, or whether you prefer this to land as experimental/unreleased first?

aannoo:
did you test it?

---

## Findings

### B1 (blocker) — no bootstrap ever reaches a launched grok agent

A launched grok agent never receives the hcom bootstrap. Not intermittently — never.

- `bootstrap.rs:491` has a grok branch (`DELIVERY_AUTO` + `GROK_DELIVERY`), so the text exists.
- Nothing delivers it. `hooks/grok.rs` never calls `common::inject_bootstrap_once`; every other
  hooked tool does (`claude.rs:1768/1935`, `copilot.rs:339/484`, `cursor.rs:534`, `kimi.rs:549`,
  `gemini.rs:414`).
- `grok.rs:427` states the intent: *"Bootstrap must use launch-time channels (`--rules` / skill),
  not hook stdout."* That channel was never built — `launcher.rs` has only `build_codex_bootstrap`;
  `LaunchTool::Grok` just appends `config.grok_args`.
- The only path that emits it is `hcom start` (ad-hoc join, `commands/start.rs:515/767`).

Live proof: `grep -c "HCOM SESSION"` → **0** in both grok session transcripts. `name_announced`
never flips because nothing reads it.

**This pre-empts the PR's headline mechanism.** With no bootstrap, the agent meets an undocumented
`hcom: wake` and goes hunting:

```
[1] USER: hcom: wake
    ├─ run_terminal_command  hcom check                    (not a command)
    ├─ run_terminal_command  hcom --help | head -80; hcom status; ls ~/.hcom
    ├─ run_terminal_command  hcom list; hcom listen --help; env | grep -i hcom
    ├─ run_terminal_command  hcom listen 5 --name grokt-kumo      <-- self-pull
    └─ run_terminal_command  hcom send --name grokt-kumo '@diro GROKPONG-1'
```

5 tool calls, ~36s, for one message. `hcom listen` advances `last_event_id`, so by the time Stop
fires there is nothing pending and the Stop path has nothing to deliver. Turns 2 and 3 did the same.
Consistent with **zero `grok.stop.*` lines in the entire `hcom.log`** across all grok sessions
(July 26 + Aug 7) — the handler isn't broken, it never has work by the time it runs.

Second defect in the same area: `GROK_DELIVERY` (`bootstrap.rs:142-149`) tells the agent *"A prompt
that is only `<hcom>` is a wake trigger"*, but grok's sentinel is `hcom: wake` (`delivery.rs:396`).
The text describes a string the agent will never see — wording copied from the cursor/copilot
blocks, which correctly describe a bare `<hcom>`.

### B2 (blocker) — unguarded claude-compat `stop_failure` marks a live grok session dead and unresumable

Grok loads both hook sources, so hcom's Claude handlers run under grok. PR81 guards four sites in
`claude.rs` with `is_grok_host()` (1784, 1866, 1873, 1943) but **not `handle_stop_failure`
(`claude.rs:1692`)**.

Live consequence — on hitting the account usage limit, `claude.rs:1692` wrote
`inactive / failure:rate_limit` while the grok process was still running:

```
$ hcom list
[GROK]  ○ grokt-kumo +1  34s ago: inactive: failure:rate_limit

$ pgrep -fl grok
47319 /Users/anno/.local/bin/grok      <-- alive
47306 .../hcom pty grok                <-- alive

$ hcom r kumo
Error: No stopped snapshot found for 'kumo'.
```

No `life.stopped` event was emitted, so the agent was simultaneously alive, marked dead, and
unresumable. Only an explicit `hcom kill` restored resumability. The pending message was correctly
preserved throughout — no force-ack.

**Scope of the bug is grok-specific even though the write isn't.** `claude.rs:1692` has fired for a
real claude agent (`muka`) four times in three days, all `failure:server_error`, none with a
`life.stopped`. But on claude it self-heals: the next hook flips it back within 0–60s (25956→25957
same second via `claude.rs:1967`; 26001→26002 at +60s via `claude.rs:2022`), because claude keeps
firing hooks after an API failure. Grok does not — after `failure:rate_limit` it sat on the modal
usage-limit dialog and stopped emitting hook events entirely (`kumo`: one stray `active/prompt` at
19:47:01, back to failure at 19:47:05, then silence; same on the resumed session). Terminal, not
transient.

So: "inactive without a `life.stopped` snapshot" is a pre-existing generic gap and **not** PR81's —
it is benign everywhere else because every tool that hits it self-heals. The missing
`is_grok_host()` guard at `claude.rs:1692` is the merge-blocking part.

### M1 — cold-start double wake produces a spurious extra turn

First delivery double-submitted the sentinel:

```
19:41:57  delivery.grok_force_enter       Forcing Enter after wake sentinel (bytes=10)
19:41:59  delivery.grok_retry_enter       still pending after 1.24s; re-Enter (attempt=2/3)
19:42:00  delivery.grok_retry_enter       still pending after 1.23s; re-Enter (attempt=3/3)
19:42:01  delivery.grok_wake_unconfirmed  unconfirmed after 3 Enters; leaving pending
19:42:01  [UPS fires]                     grok.rs:455 → active/prompt        <-- one beat late
19:42:02  delivery.grok_force_enter       2nd sentinel typed
19:42:02  delivery.grok_wake_done         enter_attempt=1, turn_started=true
```

`GROK_SUBMIT_CONFIRM` (1200ms, `delivery.rs:1117`) × `MAX_ENTER_ATTEMPTS` (3, `delivery.rs:1232`)
≈ 3.7s of confirm budget, but grok's **first-turn** UPS hook latency was ~4s (hooks bind lazily on
the first turn, same as codex). The budget expires just before the hook lands, delivery falls back
to `State::Pending`, and Pending re-types the whole sentinel. The queued `hcom: wake` auto-submits
when turn 1 ends:

```
[2] USER: hcom: wake
ASSISTANT: No pending hcom messages for grokt-kumo — inbox is clear
```

Cost: a wasted turn and tokens. Deliveries 2 and 3 were clean — cold-start only.

### M2 — status→listening depends solely on Stop, and already goes stale

Two mechanisms, one conclusion.

*Observed:* turn boundaries with no `listening` written at all — grok.rs writers only:

```
gile  19:42:47  active  prompt
gile  19:43:03  active  prompt      <-- no listening between
kumo  19:45:34  active  trigger
kumo  19:47:01  active  prompt      <-- no listening between
```

`gile` was not rate-limited in that window (it kept working until 19:45:39), so quota doesn't
explain it. Most likely M1's queued second wake submitting without a Stop on the boundary —
mechanism not isolated.

*Latent:* grok caps stop-hook continuations at `MAX_STOP_HOOK_CONTINUATIONS_PER_TURN = 8`
(`stop_gate.rs:9`), and at the cap `run_stop_gate` returns `AllowStop` **before dispatching any
hook** (`stop_gate.rs:272-282`, comment: *"At the cap no hook is consulted or notified"*). No
delivery, no `ST_LISTENING`, no notify. Not exercised in testing — hcom never returned
`additionalContext` once, so the counter never left 0 — but it becomes reachable the moment B1 is
fixed and the Stop path starts carrying traffic. hcom also never reads `stopHookActive`
(`event.rs:359-362`), the field grok provides precisely to signal an in-progress continuation chain.

### M3 — `is_grok_host()` is an ambient env switch that can silently disable Claude delivery

`common.rs:81` keys on `GROK_SESSION_ID` / `GROK_HOOK_EVENT` / `GROK_HOOK_NAME` and gates Claude's
delivery paths (`claude.rs:1784`, `1873`, `1943`).

hcom-launched children are safe — `tool_detection.rs:188` clears those vars for children. But a
`claude` started from grok's own `run_terminal_command` (not via hcom) inherits `GROK_SESSION_ID`,
and that claude instance loses Stop and UPS delivery with no error and no log beyond `grok_host=true`
in `stop.enter`. Narrow, silent, cross-tool.

### M4 — the session-end Stop guard is dead code

`is_session_end_stop` (`grok.rs:489`) matches `channel_closed` / `shutdown` / `user_exit` / …, but
grok hardcodes `reason: "end_turn"` on every Stop (`stop_gate.rs:224`) — those strings belong to
`SessionEnd`, a different event. `is_cancellable_stop_status` (`grok.rs:512`) reads a `status` field
the Stop payload doesn't have (`event.rs:357`). Both branches are unreachable, and
`session_end_stop_reasons_are_detected` pins a contract grok never emits.

The scenario is real — the live transcript has a `stop` hook firing **after** `session_end`
(`updates.jsonl` idx 135→136, again 144→145). What actually prevents delivery into a dead session is
`finalize_session` → `stop_instance` deleting the bindings, so the trailing Stop can't resolve an
instance and returns early.

### L1 — `merge_grok_args`: incomplete value table, one non-existent flag

Value-taking in real `grok --help` but missing from `VALUE_FLAGS`:

- `-w` (short form of `--worktree`; only the long form is listed)
- `--agents <JSON>`
- `--json-schema <SCHEMA>`

On resume, a value not recognised as belonging to a flag falls into the "bare positional task
prompt" branch (`resume.rs:1148-1151`) and is dropped, leaving a dangling flag → grok errors or
swallows the next token. Also:

- `--ref` (`resume.rs:1083`) is not a grok flag; only `--worktree-ref` exists.
- `--worktree` takes an **optional** value (`-w, --worktree [<WORKTREE>]`), so treating it as a
  value flag lets the drop path consume the following token (`resume.rs:1141-1145` has no `is_flag`
  guard).
- No unit test for `merge_grok_args`, while `merge_copilot_args` has two (`resume.rs:3686`, `3700`).

### L2 — leftovers from the abandoned full-body-inject design

- `common.rs:93 prompt_already_carries_hcom_body` — no callers in `src/`; only its own test. `pub`,
  so no dead_code warning. Delete.
- `delivery.rs:1109-1117` — `GROK_SUBMIT_CONFIRM`'s doc comment describes *"a Grok full-body
  inject"* that *"scales with payload size … up to ~4s"*. Neither is true of the flat 1200ms
  constant it sits on; two doc comments merged during the `d482ea5f` rewrite.
- `grok.rs:404/412/429` return `json!({ "env": grok_session_env(ctx) })` from SessionStart and UPS.
  Grok's hook output schema has no `env` key and observe-event stdout is discarded regardless —
  10 env lookups plus a JSON object per hook, for nothing.

### L3 — test quality and missing integration coverage

- `stop_output_uses_additional_context_not_followup` (`grok.rs:856`) builds a `json!` literal inside
  the test and asserts on that literal. It never calls `handle_stop`; it would still pass if the
  handler switched to `followup_message` tomorrow.
- `session_end_stop_reasons_are_detected` pins a contract grok never emits (M4).
- Nothing in `tests/`: no grok case in `test_pty_delivery.rs`, no `real_tool_grok` slice.

### L4 — dual hook sets: doubled writes, one clobber, and a missing event

Every event runs two hcom processes (`global/hcom:*` native and `global/settings:*` claude-compat),
each opening the DB — ~65ms each in the live logs, ~130ms per pair. Each status transition lands
twice:

```
26232  19:42:01  active     prompt   <- grok.rs:455
26233  19:42:01  active     prompt   <- claude.rs:1944
26256  19:42:40  listening           <- grok.rs:554
26257  19:42:40  listening           <- claude.rs:1874
```

Mostly harmless duplication, except they sometimes disagree — at 19:45:34 grok.rs wrote
`active/trigger` and claude.rs immediately overwrote it with `active/prompt`, discarding the more
accurate classification.

Separately: `SubagentStop` is a Stop-class gate on grok (`event.rs:665`) but isn't in
`GROK_HOOK_COMMANDS` (`grok.rs:27-34`), so grok subagents (`spawn_subagent`) get no delivery gate.

### L5 — grok approval prompts may not set `blocked`

Grok raised a permission dialog while a message was pending (`+1`), and hcom kept the instance
`▶ active` — never `■ blocked`. Pending + approval is normally the condition that latches
`blocked(pty:approval)`, so approval detection looks unwired for grok. Not isolated.
`hcom term inject kumo '2'` approved it correctly, so the pane is drivable.

### L6 — no composer scrape

`pty/screen.rs:609` returns `None` for `Tool::Grok`, so `prompt_empty` is always false and
`input_text` always null. The spec compensates (`require_prompt_empty: false`,
`launch_requires_ready: false`, empty `ready_pattern`), and the PR acknowledges it. It is the reason
the wake has to force-Enter blind, which is what makes M1 possible.

### L7 — SKILL.md points at a reference doc that has no grok content

`skills/hcom-agent-messaging/SKILL.md` now advertises `references/cross-tool.md` as covering
*"…kimi + copilot + grok collaboration details and per-tool quirks"*, but that file was not touched:

```
$ grep -c -i grok skills/hcom-agent-messaging/references/cross-tool.md
0
```

maneeshpatel777:
https://github.com/aannoo/hcom/discussions/133

may help

lunetics:
+1

maneeshpatel777:
> +1

Thanks for the +1, I've just pushed the latest version. Like I said it's custom made for my use, but please hcom peeps go strip the underlying architecture and integrate it.

Seems to be working flawlessly on my side, I still use grok build as part of my hcom rooms/fleet.

kiala9:
@aannoo, pushed [dead09b](https://github.com/kiala9/hcom/commit/dead09b6ccedfa81389c41bc977bdf98b299b554) to update this PR.

The managed-session delivery path now differs from the older PR description: `hcom grok` uses a thin native ACP client connected to the same Grok leader/session. Messages enter Grok's native queue, without typing a wake sentinel or pressing Enter in the composer. The existing hcom mailbox remains authoritative; only a matching successful `end_turn` acknowledges delivery. Stop does not consume the ACP-owned batch a second time. Unmanaged hook-only sessions retain Stop delivery.

I reran the final hcom implementation on **unmodified official Grok 1.0.13 (5e9a58528b76), Windows**:

- Busy native turn followed by queued hcom delivery and a real `hcom send` reply recorded in hcom's database.
- Two-line draft and mid-text cursor preserved, including across a native permission dialog with one-time approval.
- Headless automatic reception, visible response, and mailbox ACK in the same session, without terminal input.
- Killed the owned ACP transport after submission and waited through the actual native Stop: unread mail retained, explicit `acp_unacknowledged`, no ACK, no automatic replay or Stop double-consumption, and unchanged draft/cursor.
- Unsupported optional allow/deny flags, aliases, and web-search restriction rejected before creating an agent, rather than silently losing restrictions.

The locked build passed. Full `cargo test --locked` passed with **2114 passed, 0 failed, 16 ignored** in a native Windows console. The existing Windows process-exit test needs Windows `timeout.exe` first in PATH and a console; Git Bash's shadowing executable and redirected stdin produced environment failures before that rerun. No production code or tests were changed to bypass it.

**No Grok repository patch or custom Grok build is part of this update or required for ordinary managed usage.** A separate native policy-inheritance experiment is deferred. Optional policy flags still require native support and fail explicitly on the tested stock version.

GitHub currently reports conflicts with `main`; this push does not resolve those or claim merge readiness. Full recovery/reclaim/cancellation acceptance remains separate from the stock-Grok paths tested above.

kiala9:
Follow-up: merged current upstream `main` (`caa1b53`) in [b961dc4](https://github.com/kiala9/hcom/commit/b961dc4b312c0736731b6b023d1a4276b0714b54), resolving the conflict without rewriting the branch history. GitHub now reports **MERGEABLE / CLEAN**.

[CI for this exact commit is green](https://github.com/aannoo/hcom/actions/runs/34737553573): all 11 jobs passed, including Linux Rust/MSRV, Windows build/tests, typecheck, dist-check, and all six real-tool matrix entries. The PR release-plan check also passed; publishing jobs were correctly skipped.

The merged code was also rechecked locally: format, all-target Clippy, build, and 2134 passing Windows tests (16 ignored), plus fresh official Grok 1.0.13 busy/reply/draft-preservation, headless, and transport-interruption runs. No native Grok repository changes or custom build requirement were introduced.

aannoo:
@kiala9 ok nice i'll look at it

PR #82, opened 2026-07-14
fix(delivery): reap stale subagent freeze so busy parents keep receiving

zane668:
## Problem

A `Task` subagent that dies without a clean `SubagentStop` leaves the parent instance's `running_tasks.active` stuck at `true`. While it is stuck, every parent `PostToolUse` short-circuits down the "in subagent context" branch and delivers nothing, so hcom messages addressed to a busy parent are withheld until the user happens to type a prompt — `UserPromptSubmit` was the only hook running the dead-subagent cleanup.

Seen in the wild as a parent whose `running_tasks` still tracked `agent_id`s that were no longer present in the `instances` table: the subagent had gone, but the parent stayed frozen and silently stopped receiving.

## Fix

On the parent's own `PostToolUse`, reconcile the freeze before honoring it: reap dead subagents (reusing `check_dead_subagents`) and re-evaluate `in_subagent_context`. A live subagent keeps its `instances` row, so it is never reaped — `active` stays `true` and parent messages remain deferred to `end_task` rather than being mis-injected into the subagent's own context. This is scoped to `PostToolUse` because that is the only suppressed hook that carries a parent message delivery.

### Follow-up: only act on definitive death signals

Reusing `check_dead_subagents` wholesale pulled the transcript-mtime staleness heuristic into the reconciliation path. A live subagent that simply hasn't written its transcript for a while (a long tool call, waiting on I/O) can cross the stale threshold without being dead; reaping it would unfreeze the parent mid-Task and let the next subagent tool call inject the parent's messages into the subagent's context — a new cross-delivery vector.

To avoid that, `DeathSignals` scopes which signals a sweep may act on:

- `All` (DB row missing + stale mtime + interrupt marker) stays in effect at `UserPromptSubmit`, preserving existing behavior there.
- `DefinitiveOnly` (DB row missing or interrupt marker) is what the new `PostToolUse` reconciliation uses; the mtime heuristic is excluded.

## Tests

Regression tests cover: stale freeze cleared and delivery resumes to the parent; a live subagent kept frozen; and a live subagent whose transcript mtime is aged past the stale threshold — `DeathSignals::All` must flag it (fixture sanity / `UserPromptSubmit` behavior preserved) while reconciliation keeps the parent frozen and `running_tasks` intact.

Verified locally with `cargo fmt --all -- --check`, `cargo clippy --all-targets --locked -- -D warnings`, and `cargo test --locked` (all green).

Only `src/hooks/claude.rs` is touched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/hooks/claude.rs
+++ b/src/hooks/claude.rs
@@ -271,10 +271,12 @@ fn route_claude_hook(
     //
     // This is scoped to PostToolUse because that is the only suppressed hook
     // that carries a parent message delivery. It is safe against genuine
-    // freezes: a live subagent keeps its `instances` row, so
-    // `check_dead_subagents` never reaps it, `active` stays true, and parent
-    // messages are still deferred to `end_task` rather than mis-injected into
-    // the subagent's own context.
+    // freezes because the reap is restricted to deterministic death signals
+    // (`DeathSignals::DefinitiveOnly`): a live subagent keeps its `instances`
+    // row and carries no interrupt marker, so it is never reaped — even if
+    // its transcript mtime crosses the stale threshold during a long tool
+    // call. `active` stays true and parent messages remain deferred to
+    // `end_task` rather than mis-injected into the subagent's own context.
     if is_in_subagent_ctx && hook_type == HOOK_POST {
         is_in_subagent_ctx = reconcile_stale_subagent_freeze(
             db,
@@ -288,7 +290,7 @@ fn route_claude_hook(
 
         if hook_type == HOOK_USERPROMPTSUBMIT {
             let transcript_path = payload.transcript_path.as_deref().unwrap_or("");
-            cleanup_dead_subagents(db, &session_id, transcript_path);
+            cleanup_dead_subagents(db, &session_id, transcript_path, DeathSignals::All);
             // Fall through to parent handler for PTY message delivery
         }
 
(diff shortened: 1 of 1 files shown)


zane668:
Heads-up for whoever picks these up 👋

The three fork PRs I opened (#82, #83, #84) all show the same red checks — `rust-tests`, `windows-build`, and `plan`. I dug into the logs, and these look like pre-existing / toolchain-drift issues on `main` rather than anything the PRs introduce (each PR is a clean single-file diff that doesn't touch the files below):

**1. `rust-tests` + `windows-build`** — both fail at the clippy step on `main`'s own code:

```
error: redundant reference in `println!` argument
  --> src/commands/list.rs:834:38
   |
834|         println!("  Time:       {}", &entry.timestamp);
   |                                      ^^^^^^^^^^^^^^^^^ help: remove the redundant `&`: `entry.timestamp`
   = note: `-D clippy::useless-borrows-in-formatting`
```

`clippy::useless_borrows_in_formatting` was tightened in Rust 1.97, and CI installs the latest stable — so the `&` on `src/commands/list.rs:834` now trips `-D warnings`. Dropping the `&` clears it. (Local clippy on 1.96 still passes, which is why it slipped through.)
Logs: [rust-tests](https://github.com/aannoo/hcom/actions/runs/29330924662/job/87078400407) · [windows-build](https://github.com/aannoo/hcom/actions/runs/29330924662/job/87078400510)

**2. `plan`** — `dist plan` reports `.github/workflows/release.yml has out of date contents and needs to be regenerated` (the newer cargo-dist no longer emits the `swatinem/rust-cache@v2` step). A `dist init` regen should clear it.
Log: [plan](https://github.com/aannoo/hcom/actions/runs/29330925286/job/87078402611)

Just flagging to save some triage time — no rush at all, and happy to rebase once `main` is green. Thanks for the project! 🙏

aannoo:
thanks, did it different, should be fixed with these:

26287c207bb7cf7bc441cb480f3ae8b60f02120b — Routes hooks by Claude agent_id instead of running_tasks.active, so an uncleanly terminated child cannot freeze parent message delivery.
9fbdd78ed78b0abd020d2ddfe9ae7b5871a48be3 — Prevents concurrent duplicate SubagentStop hooks from racing teardown and delivery.
d6b4ac52f22ce3428203b9c60cade394be807199 — Makes child teardown exactly-once and recoverable if a stop-hook process crashes.
2a841f95c2156b0530e708d82bc5d51192b39362 — Makes failed teardown transactionally retryable instead of leaving inconsistent task tracking.
38cc5fb8ac1803479df942cd3950bd10c3766053 — Verifies the exact child actor for hcom commands, preventing parent/child/sibling misattribution.

PR #88, opened 2026-07-15
fix(test): prevent relay push subprocesses in unit tests

zane668:
## Summary

- centralize the duplicated best-effort `hcom relay push` spawn sites
- compile the child-process spawn out under `cfg(test)`
- add a regression test using an executable shim and marker file

## Why

A unit-test process can spawn the real `hcom` binary after its temporary environment has been removed. The child may then fall back to the developer's real `~/.hcom`, so in-process test isolation alone cannot protect that boundary.

Production and integration-test builds retain the existing fire-and-forget relay push behavior through the `cfg(not(test))` branch.

## Test plan

- `cargo fmt -- --check`
- `HCOM_DIR=$(mktemp -d) cargo clippy --all-targets --all-features -- -D warnings`
- `HCOM_DIR=$(mktemp -d) cargo test`
- focused regression: `HCOM_DIR=$(mktemp -d) cargo test relay::tests::test_background_push_is_disabled_in_unit_tests -- --exact`

Local result: 1993 passed, 0 failed, 15 ignored.

### DIFF
--- a/src/commands/bundle.rs
+++ b/src/commands/bundle.rs
@@ -1196,15 +1196,7 @@ fn create_and_log_bundle(
     match bundles::create_bundle_event(bundle, &instance, created_by, db) {
         Ok(bundle_id) => {
             // Trigger relay push (best-effort)
-            let prefix = crate::runtime_env::get_hcom_prefix();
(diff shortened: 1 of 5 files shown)


aannoo:
thanks. made some small fixes also

PR #89, opened 2026-07-15
test: fail closed on non-temporary database paths

zane668:
## Summary

- Redirect non-temporary `HCOM_DIR` values to a process-local temp directory in unit tests.
- Fail closed when `HcomDb::open_raw` receives a filesystem path outside the system temp tree during unit tests.
- Keep raw `HCOM_DIR` resolution tests pure by calling `resolve_hcom_dir_from_env` directly; they never seed process-global config or relax the DB tripwire.
- Isolate send, delivery, and hook tests whose call paths reopen process-global hcom state.

Canonical deepest-existing-ancestor checks reject lexical temp paths that escape through symlinks.

## Why

Tests may run inside an hcom-managed session and inherit its real `HCOM_DIR`. A test that forgets to install an isolated environment can then open or migrate the user's real database.

#88 prevents detached relay-push subprocesses from escaping unit-test isolation. This PR adds the complementary in-process redirect, raw database tripwire, and targeted fixture isolation.

## Production impact

None. The directory redirect and database tripwire compile only under `cfg(test)`, and the remaining changes are confined to test modules.

## Verification

- New redirect, symlink escape, concurrent resolution/open, temp-path, and `open_raw` tripwire tests pass.
- `cargo test --locked`: 2000 passed, 0 failed, 15 ignored.
- `cargo clippy --locked --all-targets --all-features -- -D warnings`
- `git diff --check`
- Full suite run with `HCOM_DIR` pointing to a valid database in a repository-external, non-temp directory: SHA-256 and mtime unchanged.
- Independent re-review passed all seven targeted guard/tripwire/concurrency tests.

## Dependency

Depends on #88. Keep this PR in draft until #88 lands; then rebase it onto the updated main branch so only the PR-B commit remains.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,37 +7,33 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
-      - uses: dtolnay/rust-toolchain@stable
+      # Pinned so a new stable release can't fail CI on a fresh lint with no
+      # code change. Bump deliberately in its own PR. Local dev stays floating.
+      - uses: dtolnay/rust-toolchain@master
         with:
+          toolchain: "1.97.1"
           components: rustfmt, clippy
       - uses: Swatinem/rust-cache@v2.9.1
       - run: cargo fmt --all -- --check
       - run: cargo clippy --all-targets --locked -- -D warnings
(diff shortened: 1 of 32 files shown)


aannoo:
nice thanks

Issue #87, opened 2026-07-15
RFC: tracked hcom resume retains principal


zane668:
Depends on #86.

## Summary

Assuming RFC-P1's durable principal id exists, this RFC proposes that a **tracked** `hcom resume <name>` preserve the principal of that name's current or uniquely identifiable most-recently-stopped tracked lifecycle and reinject it into the new process. The lookup is lifecycle-aware; it is not an unqualified "last principal ever seen for this string" guess.

This closes the common hcom-managed resume path without changing tool-level session-id semantics. It does not generalize to adoption, fork, vanilla restart, or explicit name reclaim. Those paths mint a fresh principal.

## Problem statement

`hcom resume <name>` starts a new OS process for what the user experiences as "the same agent instance." Anything carried only in env (`HCOM_PROCESS_ID`, and `HCOM_PRINCIPAL_ID` from RFC-P1) disappears with the old process, so env alone cannot retain the principal.

But hcom's tracked resume path already preserves the registered instance name. `do_resume` (`src/commands/resume.rs:118`) resolves the argument through `resolve_display_name_or_stopped`, and the tracked branch passes `launch_name = Some(display_name)` (`src/commands/resume.rs:508-513`). RFC-P1's live instance row and durable binding record therefore provide a bounded hand-off point: use the live row when it exists; after stop has removed that row, use the uniquely newest surviving binding for that stopped lifecycle; inject that exact principal into the replacement process.

## Proposed capability

- On `hcom resume <name>`, when the resolved name corresponds to a **tracked instance lifecycle**, resolve its principal from RFC-P1 state and inject the same value into the replacement process.
- With a live row, require agreement between `instances.principal` and its binding. With no live row after stop, use only the uniquely newest surviving binding for that name; an equal-timestamp tie or inconsistent state is ambiguous and must not be retained.
- The read and resume update must be lifecycle-safe: if the row/binding is reclaimed or changes before the replacement is committed, resume must fail or mint a new principal according to the normal creation path. It must not select an older lifecycle merely because its name string matches.
- This requires no change to how any underlying tool (Claude, Codex, Gemini, etc.) manages its own session id — the mechanism rides entirely on hcom's own resume entry point and its own name-based lookup, which is tool-agnostic.

## The three-way split (must be stated honestly, no hand-waving)

hcom's own resume/fork code already distinguishes exactly the cases this RFC needs to distinguish — this is not something we're introducing, it's already there in `src/commands/resume.rs`:

[...]

Issue #85, opened 2026-07-14
hcom list --json session_id can diverge from a session's live session id (after /resume or fork)


zane668:
## Summary

For the same running instance, the `session_id` reported by `hcom list --json` can differ from that session's **live** session id (the id the underlying agent's session actually has right now). They appear to live in two different id spaces, and I'd like to understand whether that's by design before assuming it's a bug.

## What I see

- `hcom list --json <name>` (querying a **peer**, not self) reports the `session_id` that was captured when the instance registered/launched.
- After the underlying agent's session id changes during the instance's life — e.g. a Claude `/resume`, or a fork that starts a new session file — that stored value is not updated, so the instance keeps reporting its original launch-time id.
- Meanwhile the session is now actually running under a different, newer session id.

Looking at `src/commands/list.rs`, the JSON `session_id` for a looked-up instance comes from the stored `data.session_id`; only the *self*-query path overrides it with the caller's live identity (`sender_identity.session_id`). So a peer lookup always returns the registration-time value.

## Reproduction

1. Launch a session through hcom and note its `session_id` via `hcom list --json <name>`.
2. Change the underlying agent's session id mid-life — e.g. `/resume` the Claude session, or fork it, so a new session file / id is in effect.
3. `hcom list --json <name>` still reports the original launch-time `session_id`, not the id the session is currently running under.

(Most reliably observed on relay / resume / fork-style sessions, where the agent's session id legitimately changes after launch.)

## Why it matters

External integrations that correlate an hcom instance with the live agent session by `session_id` — e.g. mapping "which running session is this hcom peer" for downstream identity/permission checks — can't reconcile the two: the id from `hcom list --json` doesn't match the id the session reports about itself at runtime, so the correlation silently fails for any resumed/forked session.

## Question (intent)

Before proposing anything, I'd like to know your intent:

1. Is `session_id` in `hcom list --json` meant to be the **launch-time / PTY-bound** identity, deliberately distinct from the agent's current session-file id? Or is it intended to track the live session and simply not being updated on resume/fork?
2. If it is meant to be the launch-time id, would you be open to `hcom list --json` also exposing the **current/live** session id (an extra field), or providing some mapping between the two, so downstream tools can align on the id the session actually reports about itself?

Happy to help reproduce further or test a direction once you've weighed in on the intended semantics. I'm intentionally not proposing a fix here since it touches core identity semantics.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #91, opened 2026-07-15
fix(ci): unblock current pull request checks

zane668:
## Summary

- remove the Rust 1.97 Clippy `useless_borrows_in_formatting` failure already present on `main`
- refresh the cargo-dist 0.31.0 generated release workflow with the exact diff reported by `dist plan`

These two baseline failures currently make unrelated PRs #88, #89, and #90 red. This PR keeps the cleanup separate from those feature changes.

## Validation

- `cargo fmt --all -- --check`
- `git diff --check`
- cargo-dist plan and Rust 1.97 Clippy are revalidated by this PR CI

PR #92, opened 2026-07-15
feat(identity): add durable principal identity primitive

### LINKED ISSUE #86: RFC: durable principal ID and session mapping
## Summary

`hcom list --json <peer>` reports a launch-time `session_id` snapshot which may diverge from the tool's later session id (issue #85). This RFC proposes a separate, opaque, hcom-generated **principal id**, persists it independently of tool session ids, and exposes a principal↔current-known-session mapping.

P1 is deliberately only the identity primitive. A principal is immutable and never silently reassigned, but P1 alone does **not** decide that a new process is the same principal as an old one. A fork always receives a new principal. Retaining a principal for a tracked `hcom resume <name>` is RFC-P2. A vanilla restart outside hcom has no continuity guarantee here.

## Problem statement (in hcom's own terms)

This is a direct response to [issue #85](https://github.com/aannoo/hcom/issues/85), summarized as follows:

- `hcom list --json <name>` for a **peer** query returns the `session_id` captured at registration/launch time (`data.session_id` in `src/commands/list.rs`), sourced from `sender_identity.session_id` only on the **self** query path.
- After the underlying agent's session id changes — for example after compaction, resume, or fork — the stored value may remain the original launch-time id.
- The instance is now actually running under a different, newer session id. The two ids live in two different id spaces that never reconcile for peer lookups.
- Any downstream integration that correlates "this hcom instance" with "the live agent session" by `session_id` — e.g. for identity/permission checks — silently fails to reconcile the two once the session has resumed or forked.

The issue asks two questions of intent: (1) is the launch-time id deliberate, or an oversight; (2) if deliberate, would hcom be open to exposing the current/live session id as an extra field, or a mapping between the two.

This RFC answers question 2 with a concrete primitive: rather than making callers infer identity from whichever tool-level id appears freshest, hcom should mint and own an identifier whose value never changes or gets silently reused. The persisted mapping can then say which current hcom instance and session, if any, is associated with that principal.

Why not only refresh `session_id` on peer queries? Because `session_id` belongs to the underlying tool, and its rotation semantics differ between tools and operations. A durable identity primitive must be generated and owned by hcom. Refreshing live-session metadata can still be useful, but it is a mapping update, not an identity rule.

## Proposed capability

- **Generation**: hcom already generates a similar-shaped, self-issued, unguessable id today — `HCOM_PROCESS_ID`, a UUIDv4-like string minted once per launch (`generate_process_id`, `src/launcher.rs:469`, called at `src/launcher.rs:1807-1808` before injection into the child's environment). A principal can use the same generation pattern but has a different contract: it is persisted as the stable identity key of one admitted principal, while `HCOM_PROCESS_ID` names one OS process. The two ids must not be merged.
- **Persistence**: hcom already has a persistent mapping table for a similar-shaped problem — `process_bindings` (`process_id TEXT PRIMARY KEY, session_id TEXT, instance_name TEXT, updated_at REAL NOT NULL`, table created at `src/db/mod.rs:199`, indexed at `src/db/mod.rs:205`). A `principal_bindings` table, keyed by the durable principal id instead of the ephemeral process id, is structurally the same kind of table hcom already maintains — not a new subsystem.
- **Exposure surface** (for upstream discussion, not prescriptive):
  - `hcom list --json <name>` gains a nullable `principal` field alongside the existing `session_id`.
  - A query path (CLI and/or JSON) maps `principal → current registered instance + current-known session_id`, or explicitly returns unresolved/stopped. "Current-known" must not be presented as a guaranteed live tool session when hcom has no fresh observation.
  - Instances may receive `HCOM_PRINCIPAL_ID` at launch, mirroring `HCOM_PROCESS_ID`, so hooks can present the principal again within that admitted lifecycle.
- **Where it plugs into existing identity resolution**: hcom already has an identity-resolution priority chain (`resolve_identity_with_expectation`, `src/identity.rs:281-290`; real order: `system_sender` → `session_id` → `name` → `process_id`). A presented principal can be considered before the drifting `session_id`, provided the presentation channel is admitted and validated. The principal value alone is not authentication; exact precedence and trust rules are implementation details, not preconditions for the exposure surface.

## Cost (stated plainly, not minimized)

- **New persistence and backup obligation.** Unlike `session_id`/`process_id` bindings, which are cheaply reconstructible from live hook traffic, a principal id is **self-generated, random, and not re-derivable from anything else**. If `principal_bindings` is lost or corrupted, there is no formula to reconstruct a given principal — only partial, best-effort recovery from whatever event log entries happen to carry a copy (and only for creation paths that actually emit such an event; paths that don't, such as bare, event-less session starts, are structurally unrecoverable). This is a genuine new durability requirement, not a free addition.
- **New write path at every creation point** (launch, bare `start`, rebind, adoption resume, and fork) that must reliably persist and propagate a principal. A tracked resume is the only resume path eligible to retain an existing principal, and only if RFC-P2 is implemented; otherwise it also receives a new principal.
- **A second identifier space to reason about and document** alongside `process_id` and `session_id` — more surface for tool authors and downstream consumers to learn.

## Lifecycle boundary

| Operation | P1 result |
|---|---|
| Initial hcom launch / bare start / adoption | Mint and persist a new principal. |
| Hooks or compaction within the admitted lifecycle | Preserve the presented principal; update only observed session metadata. |
| `hcom resume <tracked-name>` | P1 alone makes no retention promise. RFC-P2 may reattach the current principal for that tracked name. |
| Fork | Always mint a new principal. A fork is a new lineage member, not continuation of the parent principal. |
| Vanilla restart outside hcom | No automatic reattachment guarantee. A later admission mechanism must prove or explicitly downgrade continuity. |
| Explicit name reclaim/rebind | Mint a new principal unless a separate, authenticated transfer operation says otherwise. Never inherit merely because the string name matches. |

## Example

Suppose hcom launches `luna` with principal `p-a`, process `proc-1`, and tool session `s-1`. A later trusted hook from the same admitted lifecycle may update the mapping to `p-a → luna / s-2`; the principal stays `p-a`. Forking `luna` creates `nova` with a different principal `p-b`. If `luna` later starts as a new process, P1 alone does not say whether it gets `p-a` or a fresh principal — tracked-resume retention requires P2, while a vanilla restart remains unresolved without a separate admission rule.

## Honest limitations

- P1 does **not** make resume, fork, or vanilla restart identity-stable. It defines, persists, and exposes the principal and mapping needed for later continuity rules.
- `current-known session_id` is only as fresh as hcom's latest trusted observation. If freshness cannot be established, the mapping must say unresolved/stale rather than claim liveness.
- Corruption/loss handling must fail closed: if persisted state is unavailable or inconsistent, report the principal as unresolvable rather than guess or redirect to another instance.
- The principal is an opaque correlation key, not proof of a human operator's identity and not, by itself, an authorization policy.

## Acceptance criteria

1. Every creation path either persists a newly minted principal or explicitly reports that principal support is unavailable; no successful tracked instance silently has two principal values.
2. `hcom list --json <name>` exposes the instance's principal without changing the meaning of the existing `session_id` field.
3. A principal lookup returns its current registered instance and current-known session id, or an explicit stopped/unresolved result. It never guesses from name or recency.
4. Repeated hooks within one admitted lifecycle do not rotate the principal, even if observed tool session metadata changes.
5. Fork, adoption, and explicit name reclaim each mint a principal different from the source/previous lifecycle.
6. Removing or corrupting the mapping cannot cause another live instance to be returned for that principal; the result is unresolved and observable.
7. Tests cover principal uniqueness, idempotent same-lifecycle writes, fork non-inheritance, name-reclaim non-inheritance, stale mapping behavior, and JSON compatibility for consumers that ignore the new field.

## Relationship to RFC-P2

RFC-P2 depends on P1 and adds one narrowly scoped continuity rule: a tracked `hcom resume <name>` retains the principal currently attached to that tracked instance. P1 remains independently useful as the durable identifier and mapping primitive, but must not be described as solving cross-process continuity by itself.

zane668:
## Summary

- add the v18 durable principal schema and exact principal-to-instance binding primitive
- assign a fresh principal on every successful instance creation path and expose exact lookup/list fields
- fail closed on torn or conflicting claims, and roll back provisional principals when a launch never becomes live
- preserve existing v17/v18 data and keep older v18 binaries able to read and use the database

## Scope

This is the P1 data and creation layer from #86. It intentionally does not add send authorization, vanilla re-attachment, tracked-resume retention, or stop semantics. Those remain separate follow-ups, including #87.

The branch is based on current main. Existing upstream CI baseline failures are isolated in #91; this PR does not duplicate those unrelated fixes.

## Verification

- cargo test --locked: 2013 passed, 0 failed, 15 ignored
- cargo fmt --check
- cargo clippy --all-targets -- -D warnings
- v17 to v18 in-place migration and existing v18 principal-row preservation
- hcom 0.7.23 binary against a new v18 database: list, send, and events pass without changing principal rows
- failure injection for process binding, backend launch, principal creation, and conflicting/missing binding states

Closes #86

### DIFF
--- a/src/commands/list.rs
+++ b/src/commands/list.rs
@@ -170,6 +170,24 @@ pub fn cmd_list(db: &HcomDb, args: &ListArgs, ctx: Option<&CommandContext>) -> i
                         );
                         return 1;
                     }
+                    Ok(crate::db::PrincipalLookup::MissingBinding { claiming_instances }) => {
+                        let payload = serde_json::json!({
+                            "name": null,
+                            "principal": target,
+                            "session_id": null,
+                            "status": "unresolved",
+                            "reason": "missing_binding",
+                            "claiming_instances": claiming_instances,
+                        });
+                        if json_output {
+                            println!("{}", serde_json::to_string(&payload).unwrap_or_default());
+                            return 0;
+                        }
+                        eprintln!(
+                            "Principal {target} is unresolved (binding missing; instance claims are diagnostic only)"
+                        );
+                        return 1;
+                    }
                     Ok(crate::db::PrincipalLookup::Unknown) if principal_target.is_some() => {
                         eprintln!("Error: unknown principal: {target}");
                         return 1;
--- a/src/commands/list.rs
+++ b/src/commands/list.rs
@@ -26,6 +26,9 @@ pub struct ListArgs {
     pub name: Option<String>,
     /// Field to extract (used with name)
     pub field: Option<String>,
+    /// Look up the exact instance recorded for a durable principal ID
+    #[arg(long, value_name = "ID", conflicts_with = "name")]
+    pub principal: Option<String>,
     /// Show recently stopped agents
     #[arg(long)]
     pub stopped: bool,
@@ -99,7 +102,8 @@ pub fn cmd_list(db: &HcomDb, args: &ListArgs, ctx: Option<&CommandContext>) -> i
     let names_output = args.names;
     let sh_output = args.sh;
     let format_template = args.format.clone();
-    let target_name = args.name.as_deref();
+    let principal_target = args.principal.as_deref();
+    let target_name = principal_target.or(args.name.as_deref());
     let field_name = args.field.as_deref();
 
     // Resolve current instance identity
@@ -134,11 +138,49 @@ pub fn cmd_list(db: &HcomDb, args: &ListArgs, ctx: Option<&CommandContext>) -> i
             return 1;
         }
 
+        let mut queried_principal = None;
         let lookup_name = if is_self {
             current_name.clone().unwrap_or_default()
         } else {
-            let resolved = resolve_display_name(db, target);
-            resolved.unwrap_or_else(|| target.to_string())
+            let resolved = principal_target
+                .is_none()
+                .then(|| resolve_display_name(db, target))
+                .flatten();
+            if let Some(name) = resolved {
+                name
+            } else {
+                match db.lookup_principal(target) {
+                    Ok(crate::db::PrincipalLookup::Resolved { instance_name, .. }) => {
+                        queried_principal = Some(target.to_string());
+                        instance_name
+                    }
+                    Ok(crate::db::PrincipalLookup::Unresolved { instance_name }) => {
+                        let payload = serde_json::json!({
+                            "name": instance_name,
+                            "principal": target,
+                            "session_id": null,
+                            "status": "unresolved",
+                        });
+                        if json_output {
+                            println!("{}", serde_json::to_string(&payload).unwrap_or_default());
+                            return 0;
+                        }
+                        eprintln!(
+                            "Principal {target} is unresolved (recorded instance: {instance_name})"
+                        );
+                        return 1;
+                    }
+                    Ok(crate::db::PrincipalLookup::Unknown) if principal_target.is_some() => {
+                        eprintln!("Error: unknown principal: {target}");
+                        return 1;
+                    }
+                    Ok(crate::db::PrincipalLookup::Unknown) => target.to_string(),
+                    Err(error) => {
+                        eprintln!("Error: principal lookup failed: {error}");
+                        return 1;
+                    }
+                }
+            }
         };
 
         if lookup_name.is_empty() {
@@ -150,6 +192,7 @@ pub fn cmd_list(db: &HcomDb, args: &ListArgs, ctx: Option<&CommandContext>) -> i
             Ok(Some(data)) => {
                 let mut payload = serde_json::json!({
                     "name": lookup_name,
+                    "principal": queried_principal.as_deref().or(data.principal.as_deref()),
                     "session_id": data.session_id,
                     "status": data.status,
                     "directory": data.directory,
@@ -181,6 +224,7 @@ pub fn cmd_list(db: &HcomDb, args: &ListArgs, ctx: Option<&CommandContext>) -> i
                 if is_self {
                     let payload = serde_json::json!({
                         "name": lookup_name,
+                        "principal": null,
                         "session_id": sender_identity.as_ref().and_then(|id| id.session_id.as_deref()).unwrap_or(""),
                     });
                     if let Some(field) = field_name {
@@ -250,6 +294,7 @@ pub fn cmd_list(db: &HcomDb, args: &ListArgs, ctx: Option<&CommandContext>) -> i
                 "unread_count": unread_counts.get(&data.name).copied().unwrap_or(0),
                 "headless": data.background != 0,
                 "session_id": data.session_id.as_deref().unwrap_or(""),
+                "principal": data.principal,
                 "directory": data.directory,
                 "parent_name": data.parent_name,
                 "agent_id": data.agent_id,
--- a/src/commands/resume.rs
+++ b/src/commands/resume.rs
@@ -3126,6 +3126,71 @@ mod tests {
     }
 
     #[test]
+    fn tracked_fork_route_mints_a_principal_distinct_from_parent() {
+        let db = test_db();
+        let mut data = serde_json::Map::new();
+        data.insert("session_id".into(), json!("session-123"));
+        data.insert("tool".into(), json!("codex"));
+        data.insert("status".into(), json!("listening"));
+        data.insert("created_at".into(), json!(1.0));
+        db.save_instance_named("luna", &data).unwrap();
+        db.create_principal_binding("p-parent", "luna").unwrap();
+
+        let plan = prepare_resume_plan(&db, "luna", true, &[], &GlobalFlags::default()).unwrap();
+        let launch = prepare_launch_for_execution(&db, &plan).unwrap();
+        let child_name = launch.name.expect("tracked fork reserved child name");
+        let mut child_env = std::collections::HashMap::new();
+        let child_principal =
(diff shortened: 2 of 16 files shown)


aannoo:
@zane668 Thanks for the pr but I don't get like what's the point of the principal identity thing - isn't it the same as what the name is? what do you need it for?

---

AI review: 
# Review — PR #92: durable principal identity (P1)

`feat(identity): add durable principal identity primitive` · zane668 · +1275/−57
Closes #86 (RFC). Stacked follow-up: #93 (retains principal on tracked resume, closes #87). Motivated by #85 (stale `session_id` on peer queries).
Reviewed at `pull/92/head`. Findings are from code reading; cargo not re-run.

## Verdict

**Do not merge — close with #86/#87.** Even with the bugs below fixed, this is 1275 lines of permanent, non-reconstructible data machinery with no runtime consumer. Its cost is front-loaded and irreversible; its only present capability is one inspection field. The problem that actually prompted it (#85) is fixable directly without any of this.

If a concrete consumer that needs exact-incarnation identity is already committed and imminent, "fix then merge" becomes defensible. Absent that on the roadmap, the sequencing is inverted: build the primitive after the thing that needs it, not before.

## Why not, even bug-free

- **Cost is permanent and front-loaded.** A v18 schema migration that can't be cleanly undone; a `principal_bindings` table that is self-generated and *not re-derivable from anything* (so it becomes a backup/retention obligation forever); a second identity namespace to document alongside `process_id`/`session_id`; and a mint invariant smeared across five creation paths (launcher / `start` / `instance_binding` / `pidtrack` / hooks) that the next path will also forget — findings 1 and 3 below are that failure mode already happening.
- **Benefit today is zero.** `principal` drives nothing — not send, routing, stop, kill, or authorization. The whole consumer surface is `hcom list --principal` (`src/commands/list.rs`). The "which incarnation was this" question it answers is already answerable from `(device_id, creation_event_id)` or the name.
- **Sequencing is inverted.** #85's real pain is that `list --json <peer>` returns a launch-time `session_id` snapshot that drifts. That's fixable directly (expose last-observed value + freshness/source) with no new namespace. PR #92 builds the #86 foundation perfectly before the consumer exists, and #93/#87 stack more on top of it.

## If it is fixed anyway — three blocking findings

All three confirmed at source.

### 1. [P1] Recreated instance rows lose their principal, unrepairably
`src/instance_binding.rs:372` `recreate_instance_from_placeholder` rebuilds the canonical row via `initialize_instance_in_position_file` with **no principal** → `instances.principal = NULL`. Called on both rebind paths in `bind_session_to_process` (`:460` canonical-exists, `:559` restore-stopped). The SessionStart hook (`src/hooks/claude.rs:599`) invokes it and mints nothing afterward for regular agents (`ensure_subagent_principal` is subagent-only).

After a stopped-session resume or session-switch merge, the live row has `principal = NULL` while the durable `principal_bindings` row survives — a torn state that is **stuck**: `create_principal_binding` (`src/db/principals.rs:51`) bails when a binding exists but the column disagrees, so nothing repairs the column later. `list --principal` then reports the live instance as principal-less. Breaks the stated invariant that every creation path persists a principal.

Reachability caveat: the tracked-resume-same-name case short-circuits at `:411` and is fine; the gap is on restore-after-stop and session-switch paths.

### 2. [P1] Concurrent v17→v18 migration can archive freshly-written v18 data
The v18 migration SQL (`src/db/mod.rs:53-63`) is **not idempotent** — bare `ALTER TABLE instances ADD COLUMN principal`, `CREATE TABLE principal_bindings`, `CREATE INDEX`, none with `IF NOT EXISTS` (contrast `init_db` at `:226-233`, which guards all three). `try_apply_migrations` (`:657`) runs a deferred transaction and never re-reads `user_version`.

Race: processes A and B both read v17; A migrates and commits v18; B — already past its `check_schema_compat` v17 read — runs `try_apply_migrations(17)`, hits `duplicate column name: principal` → `Err` → `ensure_schema` (`:393-405`) **falls straight through to snapshot + archive** with no schema re-check between the failed migration and the archive. B archives A's valid v18 DB, destroying the non-reconstructible `principal_bindings`. Realistic in hcom, where many agents/hooks open the DB at once on first upgrade. Fix: idempotent v18 statements **and** re-check `user_version` immediately before archiving.

### 3. [P2] `ensure_subagent_principal` trusts the column, not a Resolved lookup
`src/hooks/claude.rs:1665` returns `Ok(())` as soon as `principal_for_instance` finds a non-empty `instances.principal` column — it never confirms `principal_bindings` agrees. In a torn state (e.g. from finding 1, or corruption) it treats a claim-without-binding as satisfied and skips minting. Same shape in the orphan path (`src/commands/start.rs:373`). Lower severity because `create_principal_binding` writes column+binding atomically, so they can't diverge under normal operation — this is fail-closed hardening; gate on `lookup_principal(...) == Resolved`.

## Secondary design notes

- **Doesn't fix #85.** PR #92 adds a parallel id rather than refreshing `session_id`; #85 remains open and should be solved on its own.
- **"Fork" is underspecified.** The RFC's "every fork gets a new principal" means `hcom fork` (new process). In-tool native forks (codex `/fork`, omp `/branch`) rotate the session within one process/instance and keep the principal. Not a code bug, but the term needs pinning down before the model is called complete — and #93/#87 layer retention semantics on the same ambiguity.
- **Central admission gate.** Minting scattered across five paths is exactly what produced findings 1 and 3. A single validated admission gate (create + bind + rollback in one place) would prevent the next path from omitting the invariant.

PR #93, opened 2026-07-15
feat(resume): retain principal on tracked resume

### LINKED ISSUE #87: RFC: tracked hcom resume retains principal

zane668:
Implements the tracked-resume portion of #87.

Depends on #92. Until #92 lands, this stacked draft intentionally includes the durable principal foundation commits; after #92 merges it will be rebased so only the tracked-resume delta remains.

Scope:
- retain the same principal only for tracked hcom resume
- keep fork, adoption, vanilla resume, and explicit name reclaim fresh
- require lifecycle-consistent stopped evidence for row-gone recovery
- reject missing, split, tied, or raced identity evidence instead of guessing
- bind resume reservations to the launching process so stale cleanup cannot claim or delete a later lifecycle

Validation:
- 1999 unit tests and 22 CLI tests pass
- focused resume and deterministic interleaving regressions pass
- fmt, clippy with warnings denied, and diff checks pass
- the old v18 binary can read the new stopped-event payload and preserves durable bindings

Closes #87

### DIFF
--- a/src/commands/launch.rs
+++ b/src/commands/launch.rs
@@ -173,6 +173,7 @@ pub fn run(argv: &[String], flags: &GlobalFlags) -> Result<i32> {
             args: merged_args,
             persisted_args: None,
             prior_session_id: None,
+            retained_principal: None,
             tag,
             system_prompt,
             initial_prompt,
--- a/src/commands/list.rs
+++ b/src/commands/list.rs
@@ -170,6 +170,24 @@ pub fn cmd_list(db: &HcomDb, args: &ListArgs, ctx: Option<&CommandContext>) -> i
                         );
                         return 1;
                     }
+                    Ok(crate::db::PrincipalLookup::MissingBinding { claiming_instances }) => {
+                        let payload = serde_json::json!({
+                            "name": null,
+                            "principal": target,
+                            "session_id": null,
+                            "status": "unresolved",
+                            "reason": "missing_binding",
+                            "claiming_instances": claiming_instances,
+                        });
+                        if json_output {
+                            println!("{}", serde_json::to_string(&payload).unwrap_or_default());
+                            return 0;
+                        }
+                        eprintln!(
+                            "Principal {target} is unresolved (binding missing; instance claims are diagnostic only)"
+                        );
+                        return 1;
+                    }
                     Ok(crate::db::PrincipalLookup::Unknown) if principal_target.is_some() => {
                         eprintln!("Error: unknown principal: {target}");
                         return 1;
--- a/src/commands/list.rs
+++ b/src/commands/list.rs
@@ -26,6 +26,9 @@ pub struct ListArgs {
     pub name: Option<String>,
     /// Field to extract (used with name)
     pub field: Option<String>,
+    /// Look up the exact instance recorded for a durable principal ID
+    #[arg(long, value_name = "ID", conflicts_with = "name")]
+    pub principal: Option<String>,
     /// Show recently stopped agents
     #[arg(long)]
     pub stopped: bool,
@@ -99,7 +102,8 @@ pub fn cmd_list(db: &HcomDb, args: &ListArgs, ctx: Option<&CommandContext>) -> i
     let names_output = args.names;
     let sh_output = args.sh;
     let format_template = args.format.clone();
-    let target_name = args.name.as_deref();
+    let principal_target = args.principal.as_deref();
+    let target_name = principal_target.or(args.name.as_deref());
     let field_name = args.field.as_deref();
 
     // Resolve current instance identity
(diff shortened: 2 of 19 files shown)


zane668:
CI triage: the three failing jobs are the same upstream-main baseline failures isolated in #91, not P2 regressions. Both rust-tests and windows-build stop on the pre-existing redundant `&entry.timestamp` clippy error; release plan stops because the generated cargo-dist workflow is stale. All six real-tool jobs and typecheck passed on this stacked draft. After #91 and #92 land, this branch will be rebased and rerun as the single P2 delta.

PR #95, opened 2026-07-15
fix(launch): preserve authored reply semantics across launch and resume

zane668:
## Summary

- stop appending a hardcoded reply-handoff footer to CLI launch prompts
- apply the same rule to tracked/adoption resume and fork paths for every supported tool
- preserve the shared suffix helper as an explicit opt-in for callers that intentionally request it

Dispatcher-authored prompts already define whether and where a spawned session should report. Appending `When done, send your result back to @<launcher>` could contradict an explicit different report target or a no-report instruction, so these CLI-managed paths now preserve the prompt exactly as authored.

## Verification

- `cargo clippy --all-targets --all-features --locked -- -D warnings`
- `cargo test --locked`: 1971 unit tests and 20 CLI smoke tests passed; 0 failed
- focused regressions cover suffix opt-in, launch, tracked/adoption resume, and Claude/Codex fork behavior

## Upstream CI baseline

Current `main` has independent Rust 1.97 Clippy and cargo-dist workflow failures addressed by #91. This PR deliberately does not include those baseline changes; it should be refreshed after #91 lands.

### DIFF
--- a/src/commands/launch.rs
+++ b/src/commands/launch.rs
@@ -198,7 +198,10 @@ pub fn run(argv: &[String], flags: &GlobalFlags) -> Result<i32> {
             name: None, // --name is caller identity, not instance name
             skip_validation: false,
             terminal,
-            append_reply_handoff: true,
+            // CLI launches never append the reply footer: the dispatcher's
+            // prompt owns the reply semantics (a hardcoded footer contradicts
+            // prompts that say "report to X" or "no report needed").
+            append_reply_handoff: false,
         },
     )?;
 
--- a/src/commands/resume.rs
+++ b/src/commands/resume.rs
@@ -553,9 +553,8 @@ fn prepare_resume_plan_from_source(
             name: launch_name,
             skip_validation: false,
             terminal: launch_flags.terminal.clone(),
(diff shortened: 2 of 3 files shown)


zane668:
CI note: the PR-specific local validation is fully green (fmt, clippy with -D warnings, 1971 unit tests, and 20 CLI tests). The `plan`, Linux `rust-tests`, and `windows-build` failures are the known upstream-main baseline addressed by #91 (`dist` manifest drift and Rust 1.97 `useless_borrows_in_formatting` in `src/commands/list.rs`). The additional Windows Codex real-tool failure is in `real_codex_approval_gate_blocks_pending_message_then_clears_on_approval`; the unrelated full-lifecycle Codex test passed, and this PR only removes launch/resume/fork reply-footer text. I attempted a failed-job rerun, but GitHub rejected it because this account lacks repository admin rights. Maintainer rerun requested.

aannoo:
what about instead of removing the feature instead have opt-out as a launch flag or a config option or just use the actual initial prompt like instead of `hcom claude --hcom-prompt "prompt"` use `hcom claude "prompt"` to circumvent the hcom identity routing part of it. Different use cases may want or not want it

PR #96, opened 2026-07-18
respect the `pane_id_env` set by the terminal preset.

1dr0z:
This updates `build_early_launch_context` to respect the `pane_id_env` set by the terminal preset.

## Problem

`hcom kill <name>` fails to close terminal panes for agents launched in a multiplexer (herdr, tmux, etc.) running inside an outer terminal (WezTerm, etc). It targets the `pane_id` of the outer terminal instead of the multiplexer's pane id.

`build_early_launch_context` resolves `pane_id` from a hardcoded list of generic terminal env vars (`WEZTERM_PANE`, `TMUX_PANE`, etc.) and the terminal wins. This fails because hcom will run the preset's close command with the terminal's pane id.

## Reproduction

Requires starting the multiplexer's server inside the terminal, so that it inherits the terminal's env. I ran the following in an opencode session.

WezTerm > herdr > hcom opencode:
```sh
$ env | grep -E "^(HERDR_PANE_ID|TMUX_PANE|WEZTERM_PANE)"
HERDR_PANE_ID=w3:pG
WEZTERM_PANE=0

$ hcom opencode | grep -E '^Names:'
Names: mini

$ hcom kill mini
{"error":{"code":"pane_not_found","message":"pane 0 not found"},"id":"cli:pane:close"}
Failed to close herdr pane 0: exit status: 1
Process terminated, but pane remains. Retry this command with approval/escalation:
herdr pane close 0
Sent SIGTERM to process group 15539 for 'mini' (pane close failed for herdr)
  To resume: hcom r mini
```

WezTerm > tmux > hcom opencode:
```sh
$ env | grep -E "^(HERDR_PANE_ID|TMUX_PANE|WEZTERM_PANE)"
TMUX_PANE=%205
WEZTERM_PANE=4

$ hcom opencode | grep -E '^Names:'
Names: tipi

$ hcom kill tipi
can't find pane: 4
Failed to close tmux-split pane 4: exit status: 1
Sent SIGTERM to process group 18363 for 'tipi' (pane close failed for tmux-split)
  To resume: hcom r tipi
Process terminated, but pane remains. Retry this command with approval/escalation:
tmux kill-pane -t 4
```

## Fix

Make `build_early_launch_context` preset-aware. The preset has already been resolved so we know the pane id env var. The generic loop becomes a fallback. Has the nice effect of resolving panes in custom terminal presets.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1071,6 +1071,38 @@ pub fn get_merged_preset_pane_id_env(name: &str) -> Option<String> {
     get_merged_preset(name).and_then(|p| p.pane_id_env)
 }
 
+/// Environment variables that identify a terminal-local pane/window/workspace.
+///
+/// New-window runner sidecars must not replay these values from the parent:
+/// the terminal backend supplies fresh identity to the child. Include both
+/// built-in detection variables and user-configured `pane_id_env` values.
+pub fn pane_identity_env_vars() -> std::collections::HashSet<String> {
+    pane_identity_env_vars_from_path(&paths::config_toml_path())
+}
+
+fn pane_identity_env_vars_from_path(
+    toml_path: &std::path::Path,
+) -> std::collections::HashSet<String> {
+    let mut vars = crate::shared::terminal_presets::TERMINAL_ENV_MAP
+        .iter()
+        .map(|(env_var, _)| (*env_var).to_string())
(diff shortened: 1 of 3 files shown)


aannoo:
thanks

PR #105, opened 2026-08-01
start: warn on unverifiable identity and echo the name last

alexandregressier:
Two coupled failure modes observed on 2026-07-31 while driving several concurrent Claude Code sessions over the bus:

1. `hcom start` prints `[hcom:<name>]` first, followed by a long bootstrap block. A caller that snapshots output with `tail -n` never sees its own name, concludes the start did not register, and runs `hcom start` again.
2. When `CLAUDE_ENV_FILE` is absent, the SessionStart hook has nowhere to append `HCOM_CLAUDE_UNIX_SESSION_ID`, so the id never reaches the CLI and the reuse check cannot fire. The repeated `hcom start` from the same session then silently mints a second identity; the first stays bound to nothing and later reports as `launch_failed`.

Together these turn a cosmetic truncation into duplicate identities on the bus.

Changes, both output-only:
- repeat `[hcom:<name>] ready` as the final line of `start`, so any tail of the output still shows the name;
- when identity is unverifiable (claude, no explicit `--as`, no session id), warn on stderr and name the rebind command (`hcom start --as <name>`) instead of guessing.

`cargo fmt --check`, `cargo clippy --release --all-targets -- -D warnings` and `cargo test --release` pass.

### DIFF
--- a/src/commands/start.rs
+++ b/src/commands/start.rs
@@ -703,6 +703,31 @@ fn start_bare(
     }
 
     // Resolve or generate name
+    //
+    // Identity for a vanilla session rides on HCOM_CLAUDE_UNIX_SESSION_ID, which
+    // the SessionStart hook appends to CLAUDE_ENV_FILE. When that file is absent
+    // the id never reaches the CLI, the reuse check above cannot fire, and a
+    // second `hcom start` from the SAME session silently mints a SECOND identity:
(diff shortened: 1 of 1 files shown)


aannoo:
@alexandregressier added a fix commit. please check if it works for you

alexandregressier:
Sounds fair to me.

PR #107, opened 2026-08-17
Add --json output mode to send

BigCatMellow:
## What

Adds an opt-in `--json` flag to `hcom send` that prints a single line of JSON on success instead of the normal human-readable feedback:

```json
{"event_id": 3, "delivered_to": ["nova"]}
```

## Why

`send_message()` already computes the event ID it just logged (`db.log_event(...)`), but the value is discarded (`let _event_id = ...`) and the function's return type only ever exposed `delivered_to`. There's currently no way for a script or another tool wrapping `hcom send` to learn which exact database event a given send became — a caller would have to guess (e.g. "read the newest event afterward"), which isn't safe under concurrency or retries.

## How

- `send_message()`'s return type changes from `Result<Vec<String>, String>` to `Result<(i64, Vec<String>), String>` so the already-computed event ID survives the call. All call sites (the one production call in `cmd_send`, plus the unit tests in this file that call `send_message()` directly) are updated to match.
- `cmd_send()` gets a new `if args.json { ...; return 0; }` branch inserted immediately before the existing `if args.quiet { ... }` check — structurally parallel to how `--quiet` already provides an alternate output mode. When `--json` is not passed, execution reaches the exact same code as before this change.
- Error paths are unaffected: `--json` doesn't change anything on a failed send (still `Error: ...` to stderr, exit 1).

## Testing

- `cargo build` — clean, no warnings.
- `cargo clippy --all-targets -- -D warnings` — clean.
- `cargo test` — 2169 passed, 3 pre-existing failures unrelated to this change (confirmed via `git stash` + rerun against the unpatched tree: `db::tests::test_open_raw_rejects_non_temp_path`, `db::tests::test_open_raw_rejects_temp_symlink_to_non_temp_path`, `paths::tests::test_is_test_temp_path_rejects_non_temp` — all fail identically without this diff, caused by running the checkout under a temp path).
- Manually verified end-to-end: sequential `send --json` calls print monotonic real event IDs, `--json` output is exactly one parseable JSON line with nothing else on stdout, `--json --quiet` together still prints exactly one JSON line, and error paths are unaffected.

Kept the diff as small as possible — one file, no unrelated changes.


aannoo:
thanks

PR #113, opened 2026-08-20
feat(terminal): add Ptyxis support

pzxy:
## Summary

- add a Linux `ptyxis` terminal preset that opens agents in a new Ptyxis window
- auto-detect Ptyxis sessions through `PTYXIS_VERSION`
- strip and capture Ptyxis identity variables consistently across launches and relay tests
- expose Ptyxis in Linux fallback selection, the TUI preset list, config validation, and documentation

Ptyxis is intentionally open-only: its CLI can open windows and tabs, but does not expose a supported pane/window ID close API for managed `hcom kill` cleanup.

## Testing

- `bash ./scripts/typecheck.sh`
- `cargo test --locked` (2175 unit tests passed; integration suites passed)
- `cargo +1.88 check --locked --all-targets` (MSRV)
- pinned Codex real-tool tests (2 passed)
- pinned Claude real-tool tests (2 passed)
- MQTT relay roundtrip test (all 14 phases passed)
- manual smoke test with `ptyxis --new-window -- bash -c 'sleep 3'`


aannoo:
thanks

PR #115, opened 2026-08-26
fix(codex): honor effective CODEX_HOME during launch

yasinozmeen:
## Summary

- resolve Codex's state directory from the effective child launch environment after `config.toml`, `~/.hcom/env`, and caller env overlays
- use that same `CODEX_HOME` for the writability preflight, hook installation/verification, hook trust scanning, and the Codex app-server probe
- add regressions proving the child override wins and hooks are written only to the effective child home

## Problem

When hcom runs inside a sandboxed parent Codex, a writable `CODEX_HOME` supplied through `~/.hcom/env` is part of the child environment. The launch preflight and hook setup previously ran before that environment was built, so they still inspected the parent process's default `~/.codex` and rejected an otherwise valid launch with `Operation not permitted`.

## Verification

- `cargo test --locked` (`2177` unit tests and `35` integration/smoke tests passed; only opt-in real-tool/PTY tests were ignored)
- live launch from a sandboxed parent using the patched binary directly, without a wrapper: the child Codex initialized from the overridden `CODEX_HOME`, bound its hooks, and delivered an hcom message back to the parent


aannoo:
thanks

PR #117, opened 2026-08-26
fix(launcher): preserve selected Node runtime precedence

### LINKED ISSUE #116: Launcher PATH can select system Node before the resolved runtime, breaking npm-backed MCP servers
## Summary

hcom's generated launch runner can put the system executable directory ahead of the selected Node runtime directory. For harnesses installed outside Node's bin directory (notably Codex installed under pnpm), npm-backed MCP servers then start with the wrong `node`/`npx` and fail during initialization.

## Current behavior

Both Unix and Windows runner construction in `src/launcher.rs` resolve PATH helper directories in this order:

```text
[tool_bin, hcom, python3, node]
```

The directories are prepended in that order. A representative Codex launch resolves these to:

```text
~/.local/share/pnpm/bin
~/.local/bin
/usr/bin
~/.nvm/versions/node/v24.15.0/bin
```

This makes bare `npx` resolve through `/usr/bin` first. MCP definitions that intentionally inherit the launch environment consequently use system Node 18 rather than the selected NVM Node 24.

Context7 4.0.3 requires Node `>=20.18.1`, and the installed Playwright core requires Node `>=20`, so both exit at startup. Codex reports only:

```text
MCP startup failed: Mcp error: 0: Client failed to connect: Connection closed
```

Python/uv and native MCP servers remain healthy, which makes this appear to be an MCP transport/configuration failure rather than PATH selection.

## Expected behavior

The Node runtime explicitly resolved by hcom should take precedence over a generic system executable directory in the generated child PATH, on both Unix and Windows.

## Proposed fix

Resolve the runtime directories as:

```text
[tool_bin, hcom, node, python3]
```

and add controlled regression coverage that models Codex under a pnpm directory, `python3` under `/usr/bin`, and Node under NVM. The test should verify ordering and directory de-duplication without depending on the ambient machine PATH.

## Evidence

- The behavior is present on current `main` (`79ebde1` at investigation time).
- npm debug logs from affected launches record Node 18/npm 9 and process exit 1.
- With the NVM directory moved ahead of `/usr/bin`, the same definitions record Node 24/npm 11 and exit 0.
- The affected servers are Context7 and Playwright; uv/native MCP servers continue to initialize.

I found no existing hcom issue or pull request matching the Node/PATH/MCP failure. PR #115 changes effective `CODEX_HOME` handling but does not change PATH runtime precedence.

markov-root:
## Summary

- centralize generated-runner executable-directory ordering for Unix and Windows
- resolve the selected Node runtime before the generic `python3` system directory
- preserve existing first-seen directory de-duplication
- add a controlled regression modeling Codex under pnpm, Node under NVM, and Python under the system directory

Closes #116.

## Problem

The runner previously collected directories as:

```text
[tool_bin, hcom, python3, node]
```

For Codex installed under pnpm, `python3` commonly resolves to `/usr/bin` while the selected Node runtime is under NVM. The generated child PATH therefore placed `/usr/bin` ahead of the NVM directory, so bare `npx` used system Node 18. npm-backed MCP servers requiring Node 20+ (Context7 and Playwright here) exited during initialization.

The new shared helper uses:

```text
[tool_bin, hcom, node, python3]
```

Both runner implementations call the same helper so their precedence and de-duplication rules cannot silently diverge.

## Verification

Automated:

- `cargo test --locked` (2,174 unit tests plus integration/smoke tests passed; opt-in real-tool/PTY tests remained ignored)
- focused regression: `launcher::tests::test_runner_binary_dirs_prioritize_selected_node_and_deduplicate`

Live, using this branch's debug binary to launch a fresh Codex session through hcom in a Herdr pane:

```text
Generated runner PATH:
.../target/debug
.../.local/share/pnpm/bin
.../.local/bin
.../.nvm/versions/node/v24.15.0/bin
/usr/bin

command -v node -> .../.nvm/versions/node/v24.15.0/bin/node
command -v npx  -> .../.nvm/versions/node/v24.15.0/bin/npx
node --version  -> v24.15.0
npm --version   -> 11.12.1
```

From that launched Codex session:

- Context7 `query_docs` returned Node.js documentation successfully.
- Playwright `browser_navigate` loaded a controlled `data:` page and returned the expected title/snapshot successfully.
- The test pane was then killed and removed.


markov-root:
CI note: every check passed except **Real tool (codex, windows-latest)**. GitHub's only annotation is:

> The job has exceeded the maximum execution time of 25m0s

The job was cancelled while the PowerShell `cargo test --locked --test real_tool_codex ...` step was still marked `in_progress`; it produced no assertion failure, panic, or diagnostic artifact. Linux Codex, the Windows build, Windows Claude, and Windows relay all passed. I cannot rerun the upstream Actions job with my read-only repository permission (the rerun API returns 403). Could a maintainer rerun just the failed job? If it hangs again, I will treat it as a reproducible Windows Codex lifecycle issue and investigate that path.

aannoo:
thanks

PR #123, opened 2026-09-02
fix(transcript): report unavailable model history honestly

swaynehales:
## Summary

- stop reporting transport history as empty when model transcript history is unavailable
- direct operators to `hcom events` for transport-level evidence

## Verification

- `cargo clippy --all-targets --all-features --locked -- -D warnings`
- `cargo test --all-targets --locked` — 2,211 passed; 16 ignored


aannoo:
thanks

PR #125, opened 2026-09-04
fix(send): resolve remote reply origin from relay metadata

### LINKED ISSUE #124: Remote reply-to device IDs can resolve an unrelated local event
## Summary

Remote message IDs are displayed as `<origin-id>:<device-short-id>`, for example `42:BOXE`. Passing that token back through `hcom send --reply-to 42:BOXE` currently strips the device suffix and resolves `42` as a local `events.id`.

Relay imports do not preserve the origin ID as the local row ID. They allocate a new local ID and store the origin under `_relay.id` together with the source device metadata. Consequently, a remote reply can be rejected or linked to an unrelated local event when IDs differ or collide.

Observed against `v0.7.25` / `79ebde134c4d29b5ba64e5c9839a12bedb7ee125`.

## Relevant code

- `src/commands/send.rs`: remote-looking reply tokens are parsed and the numeric component is queried as a local event ID.
- `src/relay/pull.rs`: imported events preserve their origin in `_relay` metadata while receiving a new local row ID.
- `src/messages.rs`: remote-origin IDs are intentionally presented as `<id>:<device>`.

## Expected behavior

For a reply token containing a device suffix, resolve the imported event using both the relay-origin ID and an unambiguous source-device identity, then use the corresponding local event ID for reply linkage.

Resolution should fail closed when the device suffix is unknown or ambiguous. Local reply tokens without a suffix should retain their current behavior.

## Regression coverage

Please cover at least:

- an imported origin event `42:BOXE` whose local row ID is not `42`;
- a colliding unrelated local event with ID `42`;
- two source devices with the same origin event ID;
- unknown and ambiguous device suffixes;
- unchanged local-only reply behavior.

## Non-goals

- Changing the cooperative local/relay trust model.
- Treating sender labels as authorization principals.
- Changing thread inheritance or local reply semantics.

marcelokarval:
## Summary

- resolve remote `--reply-to <origin>:<SHORT>` tokens against imported relay metadata instead of unrelated local event IDs
- enforce the canonical four-character uppercase device suffix and strict integer origin metadata, failing closed on missing or ambiguous matches
- cover local-ID collisions, device ambiguity, command effects, thread inheritance, ack-loop protection, and the end-to-end CLI path

## Verification

- `cargo test --locked` (2184 passed, 1 ignored; CLI smoke 24 passed)
- focused final-SHA proof: unresolved command effect 1/1, resolver 8/8, remote CLI smoke 1/1
- independent Codex Terra/Medium adversarial rereview: PASS, no P0-P3 findings

Fixes #124

### DIFF
--- a/src/commands/send.rs
+++ b/src/commands/send.rs
@@ -476,23 +476,53 @@ pub fn send_message(
     Ok(delivery.delivered_to)
 }
 
-/// Resolve reply_to to local event ID. Returns None if not found.
+/// Resolve reply_to to local event ID. Returns None if not found or ambiguous.
 fn resolve_reply_to_local(db: &HcomDb, reply_to: &str) -> Option<i64> {
-    // reply_to can be "42" or "42:BOXE" (remote)
-    let local_part = reply_to.split(':').next()?;
-    let id: i64 = local_part.parse().ok()?;
+    let reply_to = reply_to.trim();
+    if reply_to.is_empty() {
+        return None;
+    }
 
-    // Verify event exists and is a message
-    let exists: bool = db
-        .conn()
-        .query_row(
-            "SELECT 1 FROM events WHERE id = ? AND type = 'message'",
-            rusqlite::params![id],
-            |_| Ok(true),
-        )
-        .unwrap_or(false);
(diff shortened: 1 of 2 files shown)


aannoo:
nice

PR #129, opened 2026-09-04
fix(transcript): make prefix resolution literal and unambiguous

### LINKED ISSUE #128: Transcript prefix fallback can select the wrong instance
## Problem

`resolve_instance_transcript` falls back to:

```sql
SELECT name, transcript_path, tool, session_id
FROM instances
WHERE name LIKE ?
  AND transcript_path IS NOT NULL
  AND transcript_path != ''
LIMIT 1
```

with `format!("{}%", name)`.

SQLite therefore interprets literal user characters `_` and `%` as wildcards. A request such as `test_` can select `testa1`, exposing the wrong agent transcript. The query also has no ordering or ambiguity policy, so multiple prefix candidates depend on row/query order.

## Expected behavior

- User-entered prefix characters are matched literally.
- An exact transcript-bearing instance remains highest priority.
- Prefix fallback is deterministic or fails closed on ambiguity.
- Stopped-instance and remote-device transcript routing remain compatible.

## Reproduction/acceptance

- RED test proves literal `test_` does not resolve `testa1`.
- `%`, `_`, and backslash escaping are covered.
- Multiple valid literal-prefix candidates cannot yield row-order-dependent selection.
- Focused and affected Rust tests pass.
- Upstream Linux/Windows CI passes.

## Scope

Dedicated SOL-004 fix only. Open PR #123 also touches `src/commands/transcript.rs`; keep this change rebase-compatible and resolve conflicts without absorbing that PR's unrelated diagnostic wording.

marcelokarval:
## Summary

- treat transcript instance prefixes as literal input instead of raw SQLite `LIKE` patterns
- resolve an active prefix only when exactly one transcript-bearing instance matches
- preserve exact active-match precedence and the exact stopped-session fallback

## Why

The previous fallback used `LIKE '<input>%' LIMIT 1`. Inputs containing `_` or `%` therefore behaved as SQL wildcards, and ambiguous prefixes selected whichever row SQLite returned first. Either path could show a different agent's transcript.

## Verification

- RED reproduced independently for literal `_`, literal `%`, and ambiguous prefixes
- boundary RED proved an exact stopped session was displaced by ambiguous active prefix candidates
- focused transcript tests: 32 passed, 0 failed
- full binary suite: 2180 passed, 0 failed, 1 ignored
- `cargo clippy --locked --bin hcom -- -D warnings`
- `cargo fmt -- --check`
- independent adversarial review: PASS

Fixes #128

### DIFF
--- a/src/commands/transcript.rs
+++ b/src/commands/transcript.rs
@@ -1376,13 +1376,32 @@ fn resolve_instance_transcript(
         return Some((name, path, tool, sid));
     }
 
-    // Prefix match
-    if let Ok((matched_name, path, tool, sid)) = db.conn().query_row(
-        "SELECT name, transcript_path, tool, session_id FROM instances WHERE name LIKE ? AND transcript_path IS NOT NULL AND transcript_path != '' LIMIT 1",
-        rusqlite::params![format!("{}%", name)],
-        |row| Ok((row.get::<_, String>(0)?, row.get::<_, String>(1)?, row.get::<_, String>(2)?, row.get::<_, Option<String>>(3)?)),
+    // Prefix match (literal matching; only an unambiguous single match returns immediately)
+    let escaped = name
+        .replace('\\', "\\\\")
+        .replace('%', "\\%")
+        .replace('_', "\\_");
+    let pattern = format!("{escaped}%");
+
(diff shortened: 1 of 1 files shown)


aannoo:
thanks

PR #131, opened 2026-09-04
fix(events): keep filtered wait active on unrelated unread messages

### LINKED ISSUE #130: Filtered events --wait exits success on unrelated unread messages
## Problem

`hcom events --wait` applies its requested filters to the lookback and new-event queries, but a separate unread-message preview branch exits with status 0 for any unread message addressed to the identified caller.

As a result, an unrelated message can satisfy a filtered wait such as a thread/text completion predicate even though the requested event never occurred. This can make orchestration scripts report false success.

## Expected behavior

- status 0 only when an event satisfies the requested filter/SQL predicate
- an unrelated unread message may wake and preview, but must not terminate the filtered wait
- previewing must not advance the delivery cursor or consume the message
- the wait must still terminate on a later matching event or timeout with status 1
- invalid SQL behavior and notify-endpoint cleanup remain unchanged

## Acceptance proof

- named filtered wait remains pending after an unrelated unread message
- the same wait exits 0 after a later matching message
- timeout remains status 1 when no match arrives
- unread preview is not emitted in a tight repeated loop
- focused and full offline suites pass

This is SOL-002 from the repository-wide audit and is intentionally isolated from #128 / PR #129.

marcelokarval:
Fixes #130.

## Summary

- keep filtered `events --wait` running when unrelated unread messages exist
- preserve the delivery cursor instead of consuming messages to suppress repeats
- emit the existing unread preview at most once per wait invocation
- retain matching-event, timeout, invalid-SQL, fallback-polling, and endpoint-cleanup behavior

The production change is local to `events_wait`: 6 insertions and 3 deletions, with no new dependency or persistent/global state.

## Regression proof

The new tests cover both failure modes:

- an unrelated unread message already exists when the wait starts
- an unrelated unread message arrives after the waiter is ready, followed by a matching event

Before the fix, both unit scenarios returned success prematurely and both CLI scenarios failed on `pending=false` / exit `0`. After the fix:

- focused unit tests: 5 passed, 0 failed
- focused CLI tests: 2 passed, 0 failed
- events module: 18 passed, 0 failed
- full locked suite: 2,214 passed, 0 failed, 16 ignored
- `cargo fmt --check`: passed
- `cargo clippy --locked --bin hcom --tests -- -D warnings`: passed

The CLI regression helper uses a bounded child guard that kills and reaps on deadline or unwind. Its deferred pipe draining is intentionally limited to these small-output scenarios.

### DIFF
--- a/src/commands/events.rs
+++ b/src/commands/events.rs
@@ -1442,4 +1442,200 @@ mod tests {
             _ => panic!("Expected Launch subcommand"),
         }
     }
+
+    fn setup_test_db() -> (tempfile::TempDir, HcomDb, HcomDb) {
+        let temp_dir = tempfile::tempdir().unwrap();
+        let db_path = temp_dir.path().join("events_wait_test.db");
+        let writer = HcomDb::open_raw(&db_path).unwrap();
+        writer.init_db().unwrap();
+        let reader = HcomDb::open_raw(&db_path).unwrap();
+        (temp_dir, writer, reader)
+    }
+
+    #[test]
+    fn test_events_wait_unrelated_unread_does_not_satisfy_filter() {
+        let (_temp, writer, reader) = setup_test_db();
+        writer
+            .conn()
+            .execute(
+                "INSERT INTO instances (name, tool, status, created_at, last_event_id) VALUES ('waiter', 'test', 'active', 1000.0, 0)",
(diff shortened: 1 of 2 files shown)


aannoo:
thanks

PR #136, opened 2026-09-09
fix(omp,pi): inject pending mail as steer, not a Yield turn

atacolak:
Omitted `deliverAs` on `sendUserMessage` starts a fresh user turn when the agent is idle. HCOM pending mail was using that idle path, so pings landed as a new Yield/user prompt instead of a steer.

This always passes `{ deliverAs: "steer" }` for OMP and Pi plugins. Idle mail then resumes via `Agent.continue()` (steer injected before the next provider call). Mid-turn mail interrupts instead of waiting as follow-up.

Pinned with `plugin_delivers_pending_mail_as_steer` source tests.

Already-running sessions keep the previously loaded extension until restart.


atacolak:
Closing: this was opened against upstream by mistake. Keeping the change on our fork (atacolak/hcom), not aannoo/hcom.

